PR #141, opened 2026-06-12
fix(core): resolve descriptor ids through active catalog

### LINKED ISSUE #136: core: i18n._ descriptor overload ignores catalog id
## Summary

`@palamedes/core` 0.6.0 exposes `PalamedesI18n._(idOrDescriptor: string | MessageDescriptor, ...)`, but the `MessageDescriptor` overload formats `descriptor.message` directly and never looks up `descriptor.id` in the active catalog.

That means descriptor maps migrated from Lingui, e.g. `const labels = { foo: msg\`Foo\` }; i18n._(labels.foo)`, remain in the source locale even after translated messages are loaded.

## Reproduction

```ts
import { createI18n } from "@palamedes/core";

const i18n = createI18n();
i18n.load("de", { greeting: "Hallo" });
i18n.activate("de");

console.log(i18n._("greeting"));
// => "Hallo"

console.log(i18n._({ id: "greeting", message: "Hello" }));
// => "Hello"
// expected: "Hallo"
```

## Expected behavior

When a descriptor has an `id`, `_({ id, message })` should use the catalog entry for `id`, with `message` as the fallback. This matches the public type shape and the Lingui migration path for `msg`/`defineMessage` descriptors.

## Observed impact

During an EAC Management Lingui to Palamedes migration, direct string-id calls translate correctly, but deferred descriptor maps used by formatting helpers stay in English.

## Environment

- `@palamedes/core`: 0.6.0
- Node: 24.x locally

swernerx:
## Summary

- resolve `i18n._({ id, message }, values)` through the active catalog when a descriptor id is present
- keep `message` as the fallback for missing catalog entries
- preserve message-only behavior for descriptors without an id

Fixes #136.

## Problem

The public `MessageDescriptor` overload accepted an `id`, but the descriptor branch formatted `descriptor.message` directly. Migrated Lingui-style descriptor maps such as `const labels = { foo: msg`Foo` }` therefore stayed in the source locale even when the active catalog contained a translation for the descriptor id.

## Fix

The descriptor overload now delegates id-bearing descriptors through `getMessage(id, descriptor)`, matching the existing string-id path and retaining the descriptor message as fallback. Descriptors without `id` still format their inline `message` directly.

## Validation

- `pnpm --filter @palamedes/core test`
- `git diff --check`

Note: pnpm emitted expected unsupported-platform warnings for optional native packages on darwin arm64; the core Vitest suite passed.

### DIFF
--- a/packages/core/src/index.test.ts
+++ b/packages/core/src/index.test.ts
@@ -24,6 +24,45 @@ describe("createI18n", () => {
     )
   })
 
+  it("resolves descriptor ids through the active catalog", () => {
+    const i18n = createI18n()
+
+    i18n.load("de", {
+      "inbox.summary": "{count, plural, one {# Nachricht} other {# Nachrichten}} fuer {name}",
+    })
+    i18n.activate("de")
+
+    expect(
+      i18n._(
+        { id: "inbox.summary", message: "{count, plural, one {# message} other {# messages}} for {name}" },
+        { count: 2, name: "Ada" }
+      )
+    ).toBe("2 Nachrichten fuer Ada")
+  })
+
+  it("falls back to the descriptor message when its active catalog is missing", () => {
+    const i18n = createI18n()
+    i18n.activate("de")
+
+    expect(
+      i18n._(
+        { id: "inbox.summary", message: "{count, plural, one {# message} other {# messages}} for {name}" },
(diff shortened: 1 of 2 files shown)


swernerx:
@greptileai

PR #142, opened 2026-06-12
fix(runtime): add request-local server i18n scope

### LINKED ISSUE #134: Next RSC server runtime getter loses initialized i18n across child component render
## Summary

In a Next.js App Router migration from Lingui to Palamedes 0.6.0, translated shared Server Components can render after the page/layout has awaited the app's request initialization, but `@palamedes/runtime` still sees no active server i18n instance.

The runtime currently exposes only a synchronous module-level `setServerI18nGetter()` hook. Using React `cache()` for request-local i18n state works in the page/layout function itself, but downstream Server Component rendering can call Palamedes macros in a render scope where that cached state is uninitialized.

## Reproduction shape

```ts
// shared server runtime boundary
const getRequestState = cache(() => ({ i18n: createI18n(), initialized: false }))

setServerI18nGetter(() => {
  const state = getRequestState()
  return state.initialized ? state.i18n : undefined
})

export const initPageI18n = async () => {
  const state = getRequestState()
  state.i18n.load(locale, messages)
  state.i18n.activate(locale)
  state.initialized = true
}
````

```tsx
export default async function LoginPage() {
  await initPageI18n()
  return <SharedAuthForm />
}

export function SharedAuthForm() {
  const passwordPlaceholder = t`Password`
  return <input placeholder={passwordPlaceholder} />
}
````

Observed during `next dev` on `/login`:

```
Error: No active server i18n instance. Configure @palamedes/runtime with setServerI18nGetter() before translated code runs.
    at AuthForm (.../packages/ui/src/auth/auth-form.tsx:305:63)
````

## Expected

Palamedes should document and/or provide a request-safe Next RSC server runtime pattern for macros used in downstream Server Components after page/layout initialization. Ideally the official Next integration would expose this boundary so app code does not need a module-global fallback that risks cross-request locale bleed under concurrency.

## Environment

- Palamedes 0.6.0
- Next.js 16 App Router
- React Server Components
- Monorepo with shared UI package consumed by two Next apps

## Related

This came up while migrating an existing Lingui RSC setup where every page/layout calls `initPageI18n()` before translated content renders.

swernerx:
## Problem

Direct Palamedes macros in downstream Next.js App Router Server Components compile to `getI18n()._()`. The existing example registered a server getter from the current initializer, which is fragile for React Server Components: child Server Components can render after the initializer has returned, so they need a request-local runtime scope instead of per-render global getter registration.

This caused the pattern reported in #134 to fail with `No active server i18n instance` when direct macros were used below the component that initialized i18n.

## Fix

- Add the Node-only `@palamedes/runtime/server` subpath.
- Implement `createServerI18nScope<T>()` with `AsyncLocalStorage` and one-time runtime getter registration.
- Expose both `scope.activate(i18n)` for framework render flows such as Next RSC and `scope.run(i18n, callback)` for tightly scoped request-handler work.
- Update the Next cookie example to initialize i18n through a server-only `cache()` helper and activate the request-local scope.
- Add a downstream Server Component probe that uses `t({ message })` directly, exercising the transformed `getI18n()` path.
- Document the official Next RSC pattern in `@palamedes/next-plugin`, `@palamedes/runtime`, and the Lingui migration guide.
- Refresh the Next example catalogs after extraction.

## Validation

- `pnpm --filter @palamedes/runtime test`
- `pnpm --filter @palamedes/runtime build`
- `pnpm check-types`
- `pnpm --filter @palamedes/example-nextjs-cookie exec tsc --noEmit -p tsconfig.json`
- `pnpm --filter @palamedes/next-plugin build`
- `pnpm --filter @palamedes/example-nextjs-cookie build`
- `node -e "import('@palamedes/runtime/server').then(...)"` from the Next example workspace
- `git diff --check`

Fixes #134

### DIFF
--- a/docs/migrate-from-lingui.md
+++ b/docs/migrate-from-lingui.md
@@ -158,6 +158,53 @@ import { setServerI18nGetter } from "@palamedes/runtime"
 setServerI18nGetter(() => getRequestScopedI18n())
 ```
 
+For Next.js App Router Server Components on the Node runtime, prefer the
+server-only helper:
+
+```ts
+// src/lib/i18n.server.ts
+import "server-only"
+
+import { cache } from "react"
+import { createServerI18nScope } from "@palamedes/runtime/server"
+import type { PalamedesI18n } from "@palamedes/core"
+
+export const serverI18n = createServerI18nScope<PalamedesI18n>()
+
+const loadActiveServerI18n = cache(async () => {
+  const locale = await resolveLocaleFromCookiesOrHeaders()
+  const i18n = await loadI18n(locale)
+  return { i18n, locale }
+})
+
+export async function createActiveServerI18n() {
+  const active = await loadActiveServerI18n()
+  serverI18n.activate(active.i18n)
+  return active
+}
+```
+
+```tsx
+// app/page.tsx
+import { t } from "@palamedes/core/macro"
+import { createActiveServerI18n } from "@/lib/i18n.server"
+
+function CheckoutTitle() {
+  return <h1>{t`Checkout`}</h1>
+}
+
+export default async function Page() {
+  await createActiveServerI18n()
+  return <CheckoutTitle />
+}
+```
+
+This follows the official RSC model: server-only modules prevent accidental
+client imports, React `cache()` memoizes work within the current request, and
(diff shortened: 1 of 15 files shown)


swernerx:
@greptileai

PR #216, opened 2026-06-24
chore(deps): adopt Ferrocat file combine workflow

swernerx:
## Dependency group
- Packages: `ferrocat` 1.2.1 -> 1.3.0, `ferrocat-po` 1.2.1 -> 1.3.0
- Lockfile: transitive `ferrocat-icu` 1.2.1 -> 1.3.0
- Why grouped: the public `ferrocat` facade and `ferrocat-po` catalog API ship the same file-combine workflow and need to be reviewed together.

## Upstream changes that matter here
- Ferrocat 1.3.0 adds the catalog file combine workflow planned in sebastian-software/ferrocat#152: https://github.com/sebastian-software/ferrocat/releases/tag/ferrocat-v1.3.0
- The new API handles input path reads, path-based format inference, format consistency checks, combine semantics, diagnostics, and atomic output replacement inside Ferrocat.
- Gaps: the release notes list the feature twice; I checked the 1.3.0 crate API and used `combine_catalog_files` / `CombineCatalogFilesOptions` directly.

## Local impact and code changes
- Replaced Palamedes-local file merge orchestration with `ferrocat::combine_catalog_files` while preserving Palamedes' two-input CLI guard.
- Removed Palamedes-local catalog file format inference, file reads, temp output path generation, replace/rename fallback logic, and stale file-merge error variants.
- Moved the Rust core surface closer to Ferrocat: `combine_catalog_files`, `CatalogFileCombineRequest`, `CatalogFileCombineResult`, `CatalogFileFormat`, and `CatalogConflictStrategy`.
- Moved the native Node/TypeScript surface in the same direction: `combineCatalogFiles`, `CatalogFileCombineRequest`, `CatalogFileCombineResult`, `CatalogFileFormat = "po" | "ndjson"`, and `CatalogConflictStrategy`.
- Updated the CLI to call the file-combine API while keeping the Git-facing command as `pmds catalog merge`; options now use `--format ndjson` and `--conflict-strategy`.
- Updated docs and tests to use Ferrocat's `ndjson` naming instead of the previous Palamedes-local public `json` alias.
- Added a Palamedes-level mixed-format regression test so future Ferrocat changes cannot silently turn `.po` + `.ndjson` inference into auto-conversion.
- Updated `FERROCAT_VERSION` so native version metadata matches the declared dependency.

## Validation
- `cargo fmt --all --check`: pass
- `cargo test -p palamedes catalog_combine --locked`: pass
- `cargo test --workspace --locked`: pass
- `cargo clippy --workspace --all-targets --locked -- -D warnings`: pass
- `pnpm format:check`: pass
- `pnpm --filter @palamedes/core-node check-native-types`: pass
- `pnpm --filter @palamedes/core-node test`: pass
- `pnpm --filter @palamedes/cli test`: pass
- `pnpm check-types`: pass
- `pnpm build`: pass

## Risk
- This intentionally changes the pre-1.0 Rust and Node file-combine API names to match Ferrocat rather than maintaining Palamedes-local mirror vocabulary. Palamedes is still effectively new and has no known external consumers, so this PR does not bump the crate/package version as a breaking release.
- The invalid-format and mixed-format error text now comes from Ferrocat instead of the old Palamedes helper. Regression tests still cover that failed inference does not replace the target file.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -199,9 +199,9 @@ checksum = "9f1f227452a390804cdb637b74a86990f2a7d7ba4b7d5693aac9b4dd6defd8d6"
 
 [[package]]
 name = "ferrocat"
-version = "1.2.1"
+version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ef826c9078b8f7ec87c105c991f0e0f4e900d5f30de5f4591829d01228ded67e"
+checksum = "b1eaa6ab44981713ec2c1fc9161eb1fdf2a6ab1d507221fdd4af946f3cc2a0cf"
 dependencies = [
  "ferrocat-icu",
  "ferrocat-po",
@@ -209,18 +209,18 @@ dependencies = [
 
 [[package]]
 name = "ferrocat-icu"
-version = "1.2.1"
+version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "72fbe26a76bec82113839d12983804bcbd23f0d16e12a8b16ddfda04dd61e4ed"
+checksum = "f16b4019d1062585a9ad5d56c90963f338181992c6d87c2c99441c41fec70dc5"
 dependencies = [
  "serde",
 ]
 
 [[package]]
 name = "ferrocat-po"
-version = "1.2.1"
+version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "57ef06cd734d91a47309b8b459c55379ca6e2882523e073d0fdff0132cffeb38"
+checksum = "6cfe0d8503d5b5eb2c20a9cb0d974ede77227705539fd4498d9debc4d86a5428"
 dependencies = [
  "ferrocat-icu",
  "icu_locale",
--- a/README.md
+++ b/README.md
@@ -206,7 +206,7 @@ driver:
 
 ```bash
 git config merge.palamedes-catalog.driver \
-  'pmds catalog merge --format=po --strategy=use-first --output %A %A %B'
+  'pmds catalog merge --format=po --conflict-strategy=use-first --output %A %A %B'
 ```
 
 For the full copy-paste path, including `.po` loading and the first translated
--- a/crates/palamedes-node/src/catalog.rs
+++ b/crates/palamedes-node/src/catalog.rs
@@ -89,7 +89,7 @@ pub struct CatalogCombineInput {
 }
 
 #[napi(string_enum)]
-pub enum CatalogCombineConflictStrategy {
+pub enum CatalogConflictStrategy {
     UseFirst,
     UseLast,
     Error,
@@ -112,7 +112,7 @@ pub struct CatalogCombineRequest {
     pub inputs: Vec<CatalogCombineInput>,
     pub source_locale: String,
     pub locale: Option<String>,
-    pub conflict_strategy: Option<CatalogCombineConflictStrategy>,
+    pub conflict_strategy: Option<CatalogConflictStrategy>,
     pub selection: Option<Either<CatalogCombineSelectionName, CatalogCombineSelectionThreshold>>,
     pub include_obsolete: Option<bool>,
 }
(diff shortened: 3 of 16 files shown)


swernerx:
Clean refactor — the file orchestration moves where it belongs (Ferrocat). +79/-356 LoC, no new features, all Cargo tests green. A few follow-ups before merging:

### Notes

**1. `_ =>` catch-all in `TryFrom<palamedes::CatalogMergeFormat>` (`crates/palamedes-node/src/catalog.rs:586-598`)**

The wildcard arm suggests `ferrocat::CatalogFileFormat` is `#[non_exhaustive]`. If Ferrocat 1.4 introduces e.g. XLIFF, the JS caller only sees `format!("{value:?}")` — the Rust variant name, which doesn't exist in the TS type. Suggestions:
- Short comment on the `_` arm capturing the strategy ("Palamedes V1 exposes only Po/Ndjson; new Ferrocat variants intentionally break here")
- Extend the error text with the expected values: `"Unsupported catalog merge file format {value:?}. Expected 'po' or 'json'."`

**2. `catalog_diagnostic_from_core` helper (`crates/palamedes-node/src/catalog.rs:743-747`)**

```rust
fn catalog_diagnostic_from_core(
    value: impl Into<palamedes::CatalogDiagnostic>,
) -> CatalogDiagnostic {
    CatalogDiagnostic::from(value.into())
}
```

Replaces three `.map(CatalogDiagnostic::from)` sites. Works via generics, but: does `CatalogUpdateResponse.diagnostics` really also yield ferrocat diagnostics now, or is the switch speculative there? If the latter: prefer separate, named helpers — the unified wrapper otherwise masks a semantic mismatch.

**3. Rust core is a breaking change — bump the Cargo version?**

`CatalogMergeRequest.strategy` → `.conflict_strategy`, plus four removed `PalamedesError` variants (`WriteFile`, `ReplaceFile`, `CouldNotInferCatalogMergeFormat`, `MixedCatalogMergeFormats`). You mention "less source-compatible" — if `palamedes` is consumed externally or published on crates.io, a major bump in `crates/palamedes/Cargo.toml` belongs in this PR.

**4. Keep a test for mixed formats**

The `MixedCatalogMergeFormats` logic now lives in Ferrocat. Recommendation: keep a small integration test on `merge_catalog_files` that pushes `.po` + `.json` without an explicit `format` — not to test Ferrocat's logic, but as a trip wire in case a future Ferrocat update silently flips the semantics there (e.g. accidental auto-conversion).

**5. Doc comment slightly misleading (`crates/palamedes/src/catalog_combine.rs:96-101`)**

```rust
/// Returns an error when paths are invalid, inputs cannot be parsed, formats
/// cannot be inferred, or the output file cannot be replaced.
```

Formally correct — but all of that is thrown by Ferrocat and just bubbled up. None of it is visible in the code anymore. Suggestion: "Returns an error from Ferrocat when …" or similar, so the next reader doesn't go hunting for the replace logic in this crate.

### Verdict

Approve once 1 + 3 are clarified. The others are polish.

swernerx:
Addressed in `828f538`.

- Clarified the non-exhaustive Ferrocat format fallback in the NAPI bridge and made the error actionable with the public `po` / `ndjson` values.
- Removed the generic `catalog_diagnostic_from_core` wrapper. `CatalogUpdateResponse` now maps Palamedes diagnostics directly; the Ferrocat-backed combine results convert through `palamedes::CatalogDiagnostic` explicitly.
- Kept this as a non-breaking-version PR intentionally: Palamedes is still pre-1.0, effectively unused externally, and the API move is now called out in the PR risk section instead of forcing a crate/package version bump here.
- Added a Palamedes-level regression test for inferred mixed `.po` + `.ndjson` formats leaving the output unchanged.
- Clarified the `combine_catalog_files` doc comment so it is clear those file/format/replace errors bubble up from Ferrocat.

Rechecked locally: `cargo fmt --all --check`, `cargo test -p palamedes catalog_combine --locked`, `cargo test --workspace --locked`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `pnpm format:check`, `pnpm --filter @palamedes/core-node check-native-types`, `pnpm --filter @palamedes/core-node test`, and `pnpm check-types`.

PR #237, opened 2026-06-26
fix(transform): preserve self-closing rich placeholders

### LINKED ISSUE #236: Preserve Lingui-style self-closing rich-text placeholders for migration parity
## Summary

This is a feature/parity request, not a runtime bug.

For Lingui migrations it would be useful if Palamedes preserved Lingui-style self-closing rich-text placeholders when a JSX child does not contribute translatable text. That keeps existing PO entries reusable and avoids otherwise unnecessary `msgid` drift during a technical migration.

## Example

Source JSX:

```tsx
<Trans>
  I agree to the{" "}
  <a href={COMMERCIAL_TERMS_URL}>
    Commercial Terms
    <ExternalLink className="inline h-3 w-3" />
  </a>
</Trans>
```

Lingui on the existing branch extracts:

```po
msgid "I agree to the <0>Commercial Terms<1/></0>"
```

Palamedes 0.7.7 currently extracts:

```po
msgid "I agree to the <0>Commercial Terms <1></1></0>"
```

The rendered translation behavior is not the concern here. The issue is message identity during migration: the self-closing `<1/>` child becomes `<1></1>`, and the surrounding whitespace differs, so existing translated PO entries cannot be reused by ID.

## Desired behavior

When a rich-text placeholder represents an empty/self-closing JSX child with no translatable text, Palamedes should either:

- serialize it as the Lingui-compatible self-closing form, e.g. `<1/>`, or
- otherwise normalize self-closing and empty open/close placeholders equivalently for message ID derivation.

The goal is migration parity with Lingui for existing catalogs, not a new runtime rendering capability.

swernerx:
Closes #236.

## Summary
- serializes empty JSX rich-text children as self-closing placeholders like `<1/>`
- trims placeholder-adjacent JSX whitespace before empty component placeholders so Lingui-style IDs stay reusable during migration
- adds Rust and TypeScript extraction/transform coverage for the Commercial Terms migration case

## Validation
- `cargo fmt --all --check`
- `cargo test -p palamedes`
- `pnpm --filter @palamedes/core-node-darwin-arm64 build`
- `pnpm --filter @palamedes/transform test`
- `pnpm --filter @palamedes/extractor test`
- `pnpm format:check`

### DIFF
--- a/crates/palamedes/src/extract.rs
+++ b/crates/palamedes/src/extract.rs
@@ -811,10 +811,13 @@ fn extract_jsx_children_as_message_with_state(
                     &element.children,
                     next_component_index,
                 )?;
-                parts.push(JsxMessagePart::ComponentPlaceholder(format!(
-                    "<{name}>{}</{name}>",
-                    inner.message
-                )));
+                let is_empty = inner.message.is_empty();
+                let value = if is_empty {
+                    format!("<{name}/>")
+                } else {
+                    format!("<{name}>{}</{name}>", inner.message)
+                };
+                parts.push(JsxMessagePart::ComponentPlaceholder { value, is_empty });
             }
             JSXChild::Fragment(fragment) => {
                 let inner = extract_jsx_children_as_message_with_state(
@@ -1272,6 +1275,23 @@ mod tests {
         .expect("messages should extract");
 
         assert_eq!(messages[0].message, "<0>A</0> and <1>B</1>");
+    }
+
+    #[test]
+    fn uses_self_closing_jsx_component_placeholders_for_empty_children() {
+        let messages = extract_messages(
+            r#"
+              import { Trans } from "@palamedes/react/macro"
+              const message = <Trans>I agree to the <a href={COMMERCIAL_TERMS_URL}>Commercial Terms <ExternalLink className="inline" /></a></Trans>
+            "#,
+            "test.tsx",
+        )
+        .expect("messages should extract");
+
+        assert_eq!(
+            messages[0].message,
+            "I agree to the <0>Commercial Terms<1/></0>"
+        );
     }
 
     #[test]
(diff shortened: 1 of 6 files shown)


swernerx (commented):
## Review

Checked out locally, read the code + surrounding context, ran the suites: `cargo test -p palamedes` (103 passing), `clippy`/`fmt` clean; `@palamedes/transform` (34) and `@palamedes/extractor` (22) passing against a freshly built binary. **No blockers — mergeable.**

### Strengths
- Extractor/transformer parity verified: both produce exactly `"I agree to the <0>Commercial Terms<1/></0>"` — the critical criterion for catalog reuse, tested in both Rust *and* TS.
- Clean tuple→struct refactor: the new trim path is gated behind `is_empty_component_placeholder()` and never fires for non-empty placeholders → no regression risk. The `is_empty` definition also holds for whitespace-only and value-only children.

### Should consider
- **Document the whitespace trade-off:** the trailing-whitespace trim before empty placeholders (`jsx_message.rs:95`) matches Lingui 1:1, but: because `clean_jsx_text` already collapses whitespace runs beforehand, `Foo <Icon/> bar` loses the space *before* the icon → `Foo<0/> bar`. Invisible for the issue's example (icon directly before `</a>`), but a real rendering change for an inline icon with text on both sides. Acceptable as a migration goal, but it should be noted as intentional Lingui parity in a code comment / the PR, otherwise it'll later be "optimized away" as a bug.

### Nice to have
- The placeholder construction is logically byte-identical in `extract.rs:814` and `transform/messages.rs:300` — this PR deepens the duplication (now branching logic in two places). A shared `component_placeholder(name, inner)` helper in `jsx_message.rs` would structurally rule out extractor/transformer divergence (→ catalog mismatch).
- Add a regression-guard test for an empty child with a trailing text sibling (`Foo <Icon/> bar`) to pin down the intentional trim behavior.

PR #238, opened 2026-06-26
fix(extractor): reject nested message macros

### LINKED ISSUE #235: Fail extraction for nested message macros
Nested i18n message macros should probably be a hard extractor error instead of producing partial catalog entries.

This came up while testing the EACManagement Lingui -> Palamedes migration. We originally filed #233 as a parity issue, but that framing was wrong: preserving Lingui's behavior here is not desirable. The better Palamedes behavior is to reject invalid composition early.

Problem example:

```tsx
<Trans>
  <Plural
    one="# contract"
    other="# contracts"
    value={contractCount}
  />{" "}
  ({capacityMW} MW) expiring within 24 months
</Trans>
```

With the current extractor this can turn into separate / placeholder-shaped messages, for example:

```po
msgid "<0></0> ({capacityMW} MW) expiring within 24 months"
msgid "{contractCount, plural, one {# contract} other {# contracts}}"
```

That is not useful for translators: the static sentence context and plural logic are split apart, so nobody can translate the full sentence correctly.

Expected behavior:

- `pmds extract` should fail when it sees message macros nested in a way that cannot produce one coherent translatable message.
- At minimum, reject `<Plural>` inside `<Trans>` and nested `<Trans>` / `<Plural>` combinations that would otherwise produce fragment messages.
- The diagnostic should include file and line/column, plus a rewrite hint.

Suggested diagnostic text:

```text
Nested i18n macro is not extractable as a single message.
Move the full sentence into <Plural> branches or use plural() so translators receive the complete sentence.
```

Suggested rewrite:

```tsx
{plural(contractCount, {
  one: `# contract (${capacityMW} MW) expiring within 24 months`,
  other: `# contracts (${capacityMW} MW) expiring within 24 months`,
})}
```

Rationale:

- This is stronger than a lint rule because extraction is the source of truth for catalogs.
- It prevents silently broken PO output in CI and in repos that do not run an ESLint/i18n composition rule.
- It gives authors immediate certainty that an extraction run did not create translator-hostile fragments.

EACManagement now has an app-side guardrail for this, but Palamedes should still reject the anti-pattern directly because the extractor can see that the composition cannot produce a meaningful translation unit.

swernerx:
Closes #235.

## Summary
- rejects nested Palamedes JSX message macros before they can produce fragment catalog entries
- reports a fatal diagnostic with source file, line, column, and rewrite guidance
- keeps batch extraction fatal for this authoring error so `pmds extract` fails instead of silently skipping the file
- adds Rust and TypeScript coverage for the nested `<Plural>` inside `<Trans>` case

## Validation
- `cargo fmt --all --check`
- `cargo clippy -p palamedes --all-targets -- -D warnings`
- `cargo test -p palamedes`
- `pnpm --filter @palamedes/core-node-darwin-arm64 build`
- `pnpm --filter @palamedes/transform test`
- `pnpm --filter @palamedes/extractor test`
- `pnpm format:check`

### DIFF
--- a/crates/palamedes/src/error.rs
+++ b/crates/palamedes/src/error.rs
@@ -157,4 +157,12 @@ pub enum PalamedesError {
         /// Value names provided by values but not referenced by the message.
         extra: String,
     },
+    /// A JSX message macro was nested inside another JSX message macro.
+    #[error(
+        "Nested i18n macro is not extractable as a single message at {location}. Move the full sentence into <Plural> branches or use plural() so translators receive the complete sentence."
+    )]
+    NestedMessageMacro {
+        /// Source location of the nested macro.
+        location: String,
+    },
 }
--- a/crates/palamedes/src/extract.rs
+++ b/crates/palamedes/src/extract.rs
@@ -10,9 +10,8 @@ use crate::jsx_message::{
 use oxc_allocator::Allocator;
 use oxc_ast::ast::{
     Argument, CallExpression, Expression, ImportDeclaration, ImportDeclarationSpecifier,
-    JSXAttributeValue, JSXChild, JSXElement, JSXExpression, JSXFragment, JSXOpeningElement,
-    MemberExpression, ObjectExpression, ObjectPropertyKind, TaggedTemplateExpression,
-    TemplateLiteral,
+    JSXAttributeValue, JSXChild, JSXElement, JSXExpression, JSXOpeningElement, MemberExpression,
+    ObjectExpression, ObjectPropertyKind, TaggedTemplateExpression, TemplateLiteral,
 };
 use oxc_ast_visit::{walk, Visit};
 use oxc_parser::Parser;
@@ -231,11 +230,11 @@ impl<'a> Visit<'a> for ExtractionVisitor<'a> {
             tag_name.as_str(),
             &["Trans", "Plural", "Select", "SelectOrdinal"],
         ) {
-            if let Some(nested) =
+            if let Some(nested_start) =
                 nested_message_macro_in_children(&it.children, self.imported_macros)
             {
                 self.fail(PalamedesError::NestedMessageMacro {
-                    location: self.location(nested.span.start as usize),
+                    location: self.location(nested_start),
                 });
                 return;
             }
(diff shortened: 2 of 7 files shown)


swernerx (commented):
## Review

Checked out locally, read the code + surrounding context, ran the suites: `cargo test -p palamedes` (103 passing), `cargo clippy --all-targets -- -D warnings` and `cargo fmt --all --check` clean. **No blockers — mergeable.**

### Strengths
- Implemented symmetrically in extractor and transformer. Importantly correct: the nested check runs *before* the Trans replacement in `transform/visitor.rs:96` — otherwise a nested `<Plural>` would be silently swallowed by the successful replace (`return` on line 149).
- Batch propagation is correctly **fatal**: `NestedMessageMacro` is matched alongside `ExplicitMessageIdsUnsupported` and returned as `Err` instead of being pushed into `failed_files` (`extract.rs:1111`). Covered by `batch_keeps_nested_jsx_message_macros_fatal`.
- `is_jsx_message_macro` resolves only through the import map → no false positives from local renames or third-party components with the same name.

### Should consider (debatable)
- **Gap with non-plain-JSXElement expressions:** `nested_message_macro_in_jsx_expression` only descends into `JSXElement`/`JSXFragment`. As a result `<Trans>{cond ? <Plural/> : null}</Trans>`, `{cond && <Plural/>}` and `Array.map` cases are **not** detected as nested macros. No silent miscompile, but for the ternary/logical case the extractor then throws the misleading `UnnamedPlaceholder` message ("stable placeholder name") instead of the targeted rewrite hint. For the stated goal of "clearly reject nested macros" this is a partial gap — at least scan ternary/`&&` operands too, or capture it as a known issue.

### Nice to have
- The four `nested_message_macro_*` helpers + location computation are near-identically duplicated in `extract.rs` and `transform/visitor.rs` (even two separate, both byte-based location implementations). A candidate for `jsx_message.rs` — a real refactor though, since `ImportedMacro` is defined twice separately (would need a small trait/closure to parametrize `imported_name`).
- Tests only cover the direct `<Trans><Plural/></Trans>` case; fragment/deep nesting and the ExpressionContainer path are handled in code but untested — one test each would be worthwhile.

PR #260, opened 2026-06-28
Make pmds a native Rust CLI

### LINKED ISSUE #257: cli: PO source references should be stable and relative for monorepo parent includes
## Summary

`pmds extract --clean` can emit machine-specific absolute paths in PO `#:` source references when a catalog includes source files outside the app/package directory.

This makes generated PO catalogs differ by local checkout path or CI workspace path. Downstream projects should not need a post-processing script to make extractor output reviewable and deterministic.

## Reproduction shape

An anonymized monorepo setup:

```text
repo/
  apps/web/
    palamedes.config.ts
    app/page.tsx
    components/local-card.tsx
    locales/en/messages.po
  packages/ui/
    src/shared-card.tsx
```

The app catalog includes both local app sources and shared workspace package sources:

```ts
import { defineConfig } from "@palamedes/config";

export default defineConfig({
  sourceLocale: "en",
  locales: ["en", "de"],
  catalogs: [
    {
      path: "locales/{locale}/messages",
      include: [
        "app",
        "components",
        "../../packages/ui/src",
      ],
    },
  ],
});
```

Running from `apps/web`:

```sh
pmds extract --clean
```

currently produces mixed references. Local app files are relative:

```po
#: app/page.tsx:4
msgid "Dashboard"
msgstr "Dashboard"
```

but files matched through the parent-directory include can be emitted as absolute machine paths:

```po
#: /Users/alice/work/repo/packages/ui/src/shared-card.tsx:12
msgid "Shared action"
msgstr "Shared action"
```

In CI the same source may become:

```po
#: /home/runner/work/repo/repo/packages/ui/src/shared-card.tsx:12
msgid "Shared action"
msgstr "Shared action"
```

## Expected behavior

PO source references should be stable and machine-independent.

For files outside the app/package directory, Palamedes should emit a deterministic relative reference instead of an absolute checkout path. For example, either of these would be acceptable as long as it is consistent:

```po
#: ../../packages/ui/src/shared-card.tsx:12
```

or:

```po
#: packages/ui/src/shared-card.tsx:12
```

The important contract is:

- no absolute local checkout paths in committed PO files
- no CI workspace paths in committed PO files
- stable ordering of multiple `#:` references for the same message
- repeated extraction from the same tree should leave catalogs byte-stable

If there are multiple reasonable roots in a monorepo, an explicit config option such as `sourceRoot` / `referenceRoot` would also work.

## Why this matters

Downstream projects often run extraction from app packages while sharing UI/components from workspace packages. Absolute `#:` references create large, noisy catalog diffs and make clean-tree validation depend on the machine path.

We removed our local PO-reference post-processing workaround because this should be extractor behavior, not a downstream cleanup step.

## Environment

- `@palamedes/cli`: 0.7.12
- `@palamedes/config`: 0.7.12
- Node: 24.x
- package manager: pnpm workspace

swernerx:
## Summary

- replace the Node.js `pmds` implementation with a native Rust CLI crate for extract, audit, report, watch, and catalog merge flows
- ship `@palamedes/cli` as an npm distribution wrapper for the same four native target platforms as `@palamedes/core-node`: macOS arm64, Linux x64 glibc, Linux arm64 glibc, and Windows x64 MSVC
- keep macOS x64, Linux musl, and Windows arm64 explicitly out of the published native target matrix for now
- handle the Windows review follow-up by preserving the `.exe` install target for the published Windows binary
- make the CLI postinstall path tolerate missing optional platform packages with an actionable warning, preserve POSIX executable bits, and smoke-test the public `@palamedes/cli` wrapper path
- make `palamedes.yaml` the documented default config format, with `palamedes.yml`, JSON, and TOML data configs supported by the Rust CLI and `@palamedes/config`
- switch documented config keys to kebab-case while accepting snake_case aliases at the loader boundary
- record the native CLI and YAML-first config decision in ADR-016

Fixes #257.

## Validation

- `cargo fmt --all --check`
- `cargo test --workspace --locked`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `env CI=true pnpm test`
- `env CI=true pnpm lint`
- `env CI=true pnpm check-types`
- `env CI=true pnpm format:check`
- `env CI=true pnpm check:release-set`
- `env CI=true pnpm build`
- `env CI=true pnpm install --frozen-lockfile`
- `node --check packages/cli/scripts/install.mjs`
- `node --check packages/cli/scripts/build-native.mjs`
- `node --check packages/cli/scripts/smoke.mjs`
- `node packages/cli/scripts/smoke.mjs`
- `pnpm install --lockfile-only --offline`
- GitHub Actions validate matrix passes on macOS, Ubuntu, Windows, and Rust after the wrapper-install follow-up


swernerx:
@greptileai

PR #270, opened 2026-06-30
docs: plan Ferrocat 2.0 migration

swernerx:
## Summary
- add an active RFC for migrating Palamedes from Ferrocat 1.3.1 to 2.0.0
- recommend PO as the default catalog format with FCL as an explicit opt-in storage format
- document API, CLI, N-API, TypeScript, docs, test, validation, and rollout implications

## Validation
- docs-only change
- git diff --cached --check before commit


swernerx (commented):
Reviewed the full RFC and verified its concrete API claims directly against the actual Ferrocat 2.0 source (`ferrocat-v2.0.0`).

## Overall

Strong RFC. The scope is large but coherent and correctly framed: a real product migration rather than a compatibility patch, with a clear default (PO stays default, FCL opt-in), clean separation from ADR-006, explicit non-goals, and a consistent chain from plan → tests → acceptance criteria. Treating Ferrocat 2.0 as the compatibility floor (no v1 loaders) is well-justified and keeps the migration small.

Almost every concrete API claim matches the real Ferrocat 2.0 code exactly:

| RFC claim | Verified in Ferrocat 2.0 |
|---|---|
| `CatalogMode::IcuPo` / `IcuFcl` | ✅ `ferrocat-po .../types.rs` (also exposes `GettextPo`) |
| `CatalogOrigin { file, scope: Option<String> }`, serialized `file#scope` | ✅ exact, incl. "not derived from `msgctxt`" |
| `MachineMetadata { lock, ai: Option<…> }` | ✅ |
| `AiProvenance { model, confidence: Option<f32> }` in `[0,1]` | ✅ |
| `CatalogMessage.obsolete: Option<ObsoleteInfo>` with `since` | ✅ |
| `obsolete-since` + age-based cleanup | ✅ `ObsoleteStrategy::DropObsoleteBefore(date)` |
| `Ndjson` removed, `Fcl` added | ✅ |
| `ferrocat-icu` 2.0.0 exists | ✅ |

## Substantive findings

**1. Name collision: `CatalogStorageFormat` already exists in Ferrocat.** The RFC proposes a *Palamedes-owned* `CatalogStorageFormat { Po, Fcl }` (Rust Core section) **and** a `CatalogFileFormat { Po, Fcl }`. Both names, with identical variants, are **already public exports of `ferrocat-po`** (`CatalogStorageFormat`, `CatalogFileFormat`, alongside `CatalogMode` and `CatalogSemantics`). The RFC doesn't mention these types already exist. Please make the decision explicit: re-export/wrap Ferrocat's types, or — if a Palamedes-owned type is wanted — name it distinctly (e.g. `PalamedesCatalogFormat`) to avoid confusion at the N-API boundary. This is the strongest clarity issue.

**2. The "fuzzy is gone" framing is too absolute and conflicts with the conversion requirement.** The RFC says fuzzy "is intentionally not part of the Ferrocat 2.0 model" and should be removed "entirely". That's true for the high-level catalog *model* — but the raw PO layer in `ferrocat-po` still **parses and serializes `fuzzy` flags** (`parse.rs`, `serialize.rs`, `merge.rs`). Meanwhile the RFC requires "fail before writing output when the source contains fuzzy entries", which is only implementable via that low-level flag access. Please reword to: "fuzzy is no longer a catalog *semantics* concept, but is still detected during PO parsing so conversion/audit can reject it." As written it reads contradictory.

**3. `--clean` never removes undated obsolete entries.** Per Ferrocat's `DropObsoleteBefore` contract: *"Undated obsolete entries are kept."* So existing obsolete entries without a `since` date are **never** removed by `--clean` (30-day cutoff) — only `--force-clean` (= `ObsoleteStrategy::Delete`) clears them. Worth stating in the RFC/docs; otherwise it will surprise users with older catalogs.

**4. Inconsistent version floor.** Summary/Goals say "migrate **from 1.3.1**", but the compare links use `ferrocat-v1.3.2...v2.0.0` (and `1.3.2` is a real tag). Please clarify the actual current version.

**5. `GettextPo` not addressed.** Ferrocat offers both `IcuPo` and `GettextPo`. The RFC maps `po → IcuPo` (matching Ferrocat's `default_mode`) — good, but it should state explicitly: "Palamedes always uses ICU semantics (`IcuPo`); `GettextPo` is intentionally not exposed." Otherwise it's unclear why the other PO variant is absent.

## Gaps / missing

**6. Palamedes' own release/versioning strategy.** This is a breaking change to Palamedes' public API (N-API enums, TS types, config `format`, removed `--format=ndjson`). The RFC says nothing about Palamedes' own SemVer bump, changeset, or release communication. Does Palamedes itself go 2.0?

**7. Migration path for existing `palamedes.yaml`.** Removing NDJSON config values breaks configs with `format: ndjson`. "Fail clearly" is stated, but no concrete error text/hint is given. Since FCL is effectively the NDJSON successor, an error suggesting `ndjson → fcl` would help.

**8. Reverse `convert` (FCL → PO).** Only PO→FCL is described. If FCL→PO is intentionally omitted (no opt-out path), say so as a non-goal rather than leaving it open.

**9. Pseudo-locale + FCL.** Pseudo-locale is mentioned in the CLI-relationship section but appears in neither the plan nor the FCL tests. Worth confirming pseudo-locale generation is format-independent.

## Clarity / tightening

- The "no real external users / compatibility floor" message is repeated ~5× (Summary, Goals, Non-Goals, Compatibility Stance, Risks). Once is enough.
- The **Tests** section and **Acceptance Criteria** overlap almost 1:1. Consider reducing each to its own concern (tests = behavior, AC = end state).
- The Breaking-Changes table is excellent — keep it as the centerpiece.

Bottom line: scope is right and the technical accuracy is high. Before implementation I'd only resolve the naming decision (#1) and the fuzzy framing (#2), and add the release strategy (#6); the rest are refinements.

PR #296, opened 2026-07-04
fix(release): build musl cdylib addon via target-scoped RUSTFLAGS

fastner:
## Problem

The 1.0.0 release Publish run failed on `publish-native (@palamedes/core-node-linux-x64-musl)`:

```
error: cannot produce cdylib for `palamedes-node v1.0.0` as the target
`x86_64-unknown-linux-musl` does not support these crate types
```

(The parallel `win32-x64-msvc` failure in the same run was an unrelated **transient** network error — `SocketError: other side closed` during a download step — and just needs a re-run.)

## Cause

`build-native.mjs` disabled `crt-static` only through trailing `cargo rustc --lib -- -C target-feature=-crt-static …`. Those flags reach the final crate too late for cargo's crate-type check, which runs in the build plan using the musl default `+crt-static` — under which cargo refuses to emit a `cdylib`.

## Fix

Move `-crt-static` (and `link-self-contained=+unwind`) into `CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_RUSTFLAGS` so the build plan sees crt-static disabled and allows the cdylib. Scoping to the musl **target** (not global `RUSTFLAGS`) keeps host build scripts and proc-macros untouched, and applying `link-self-contained=+unwind` target-wide supplies the in-tree unwinder for **every** cdylib built for musl — including `oxc_sourcemap` (confirmed to have a `cdylib` crate-type) — so musl-gcc never needs a shared `libgcc_s` the Ubuntu musl toolchain lacks.

## Verification

- The refactor keeps the non-musl path byte-identical; verified locally by building the darwin-arm64 addon successfully with the updated script.
- The musl cross-build cannot be exercised on macOS; it verifies in CI on the next Publish (or a manual run).

### DIFF
--- a/packages/core-node/scripts/build-native.mjs
+++ b/packages/core-node/scripts/build-native.mjs
@@ -108,10 +108,23 @@ if (muslNodeAddon) {
   // supplies the unwinder in-tree for every cdylib built for this target —
   // including the `oxc_sourcemap` cdylib — so musl-gcc never looks for a shared
   // `libgcc_s` that the Ubuntu musl toolchain does not ship.
+  //
+  // `panic=abort` is intentionally not forced here. napi-rs (`#[napi]`) wraps
+  // every `extern "C"` entry point in a panic guard, so panics are converted to
+  // JS errors and never unwind across the FFI boundary. The gnu, darwin, and
+  // win32 addons already build with the default unwind strategy for the same
+  // reason; scoping abort to musl only would diverge from them and force the
+  // whole musl dependency graph off the prebuilt unwind std.
+  //
+  // Prepend any inherited target rustflags so an externally provided value
+  // (e.g. CI optimisation overrides) is preserved rather than dropped.
   cargoEnv.CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_RUSTFLAGS = [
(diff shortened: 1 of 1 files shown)


fastner:
Thanks — addressed in 72d3fb6:

**1. Overwriting `CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_RUSTFLAGS`** — valid, applied. Now prepends any inherited value: `[process.env.CARGO_TARGET_X86_64_UNKNOWN_LINUX_MUSL_RUSTFLAGS ?? "", …].filter(Boolean).join(" ")`. (Nothing in this repo sets it today — only the `…_LINKER` — so it's defensive, but correct.)

**2. Dropped `-C panic=abort`** — intentional, now documented in-code. `palamedes-node` uses napi-rs (`#[napi]`), whose generated wrappers guard every `extern "C"` entry point and convert panics to JS errors, so no panic unwinds across the FFI boundary. The gnu/darwin/win32 addons already build with the default unwind strategy for exactly this reason; forcing `panic=abort` on musl only would diverge from them and push the whole musl dependency graph off the prebuilt unwind std. A genuine abort policy would belong in workspace `[profile.release] panic = "abort"`, not a musl-only rustflag.

Issue #284, opened 2026-07-03
Remix v3 support: adapter + demo — be the first i18n framework ready for Remix 3


swernerx:
## Motivation

[Remix v3](https://remix.run/) shipped its first beta on April 30, 2026. It is a ground-up rebuild — notably it **drops React entirely** and ships its own component model (a Preact fork: JSX, but no hooks, no re-render semantics, components built on web primitives like `EventTarget` with explicit updates). It runs directly on the Fetch API (`Request`/`Response`) across Node, Deno, and Bun.

The ecosystem around it is essentially empty right now. Existing i18n libraries are all coupled to React (react-i18next, Lingui's React components, react-intl), so none of them work with Remix v3's component model out of the box. Being the **first i18n framework with a working Remix v3 adapter + demo** is a strong positioning opportunity while the beta audience is actively looking for tooling.

## What Remix v3 looks like technically

- **Component model:** JSX without a virtual-DOM/hooks runtime; explicit updates instead of `useState`-driven re-renders; standard JavaScript control flow.
- **No bundler assumption:** Remix v3 is designed "with no expectation of static analysis" and uses `node --import` loaders for TS/JSX transformation instead of a Vite/webpack pipeline.
- **Routing/data:** routes export per-HTTP-method handlers (GET for loading, POST/PUT/DELETE for mutations); loaders re-run after mutations.
- **Frames:** page sections that refresh independently via server-rendered HTML fragments (HTMX-like partial updates).
- **Distribution:** a single `remix` package wrapping ~50 focused sub-packages (`fetch-router`, middleware, ui, …).

## Proposed work

1. **Spike / feasibility (first step):** scaffold `npx remix@next new`, inspect the ui package's component model and the `--import` loader chain, and confirm where our OXC transform can hook in.
2. **Build integration:** since there is no Vite pipeline by default, evaluate shipping a Node `--import` loader (or module hook) that runs the Palamedes macro transform — conceptually a sibling of `packages/vite-plugin` / `packages/next-plugin`. If a Remix v3 + Vite setup is viable, reuse `vite-plugin` as a shortcut for the demo.
3. **Runtime adapter:** a `packages/remix` (or `packages/remix-ui`) adapter analogous to `packages/react` / `packages/solid`, targeting Remix's explicit-update component model. Locale switching via full reload fits Remix's server-driven model well and matches our recommended default.
[...]

PR #350, opened 2026-07-06
[codex] Add experimental Remix v3 support

swernerx:
## Summary

Refs #284.

This moves the Remix v3 work from spike to an experimental first integration:

- documents the Remix v3 beta loader shape and feasibility findings
- adds `@palamedes/remix` with a Node `registerHooks()` load hook for JS macro transformation
- adds `@palamedes/remix/server` for request-local Palamedes i18n setup in Remix handlers
- adds a minimal `examples/remix-cookie` app that runs through `remix/node-tsx` plus the Palamedes register hook
- expands the Remix example with cookie locale switching, `Accept-Language` negotiation, and per-request catalog loading
- addresses review feedback around sourcemaps, exact beta pinning, public API shape, docs boundaries, and matrix smoke coverage
- wires the new package into Release Please, publishing, root TS references, API docs, and the example index

## Current Scope

The supported first path is Remix v3 server-loaded JavaScript/TypeScript modules using:

```sh
node --import remix/node-tsx --import @palamedes/remix/register server.ts
```

The register hook covers Palamedes JS macros such as `t`, `msg`, `plural`, `select`, `selectOrdinal`, and `defineMessage`.

The example also covers the first server runtime flow: `Accept-Language` or a
`locale` cookie resolves the active locale, catalogs are loaded before
translated code runs, and POST `/locale` sets the locale cookie for the next
render.

Rich JSX message macros remain intentionally experimental for Remix v3 because Remix's default loader lowers JSX before the Palamedes register hook sees the module.

The register hook is intentionally server-only for now. Browser-delivered Remix
v3 modules go through Remix's asset compiler, which does not currently expose a
Palamedes macro transform hook.

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm --filter @palamedes/transform... build`
- `node --import remix/node-tsx --import ./palamedes-register.mjs proof.mjs`
- temporary Remix HTTP smoke returned `x-palamedes-proof: Hello Ada from Remix 3`
- `./node_modules/.bin/oxfmt --check --ignore-path .gitignore --ignore-path .oxfmtignore docs/plans/2026-07-06-remix-v3-support-spike.md`
- `pnpm --filter @palamedes/remix test`
- `pnpm --filter @palamedes/remix exec tsc --noEmit -p tsconfig.json`
- `pnpm --filter @palamedes/remix build`
- `pnpm --filter @palamedes/example-remix-cookie build`
- `pnpm check:release-set`
- `pnpm lint`
- `./node_modules/.bin/oxfmt --check packages/remix examples/remix-cookie docs/api/remix.md docs/api/README.md examples/README.md`
- `pnpm --filter @palamedes/example-remix-cookie start`
- `curl -i -H 'Accept-Language: de' http://127.0.0.1:4060/`
- `curl -i -X POST -d 'locale=es' http://127.0.0.1:4060/locale`
- `curl -i -H 'Cookie: locale=es' http://127.0.0.1:4060/`
- `CI=true pnpm install --frozen-lockfile`
- `CI=true pnpm --filter @palamedes/remix test`
- `CI=true pnpm --filter @palamedes/remix exec tsc --noEmit -p tsconfig.json`
- `CI=true pnpm --filter @palamedes/remix build`
- `CI=true pnpm --filter @palamedes/example-remix-cookie build`
- `CI=true pnpm lint`
- `CI=true pnpm check:release-set`
- `CI=true pnpm verify:examples:smoke -- --id remix-cookie`
- `./node_modules/.bin/oxfmt --check packages/remix examples/remix-cookie docs/api/remix.md docs/plans/2026-07-06-remix-v3-support-spike.md scripts/example-matrix.mjs`

The final HTTP smokes returned:

- `200`, `x-palamedes-locale: de`, `<html lang="de">`, `Remix v3 rendert de mit Palamedes`, and `3 Plätze frei`
- `303` from POST `/locale` with `Set-Cookie: locale=es`
- `200`, `x-palamedes-locale: es`, `<html lang="es">`, and `3 asientos disponibles`

### DIFF
--- a/docs/plans/2026-07-06-remix-v3-support-spike.md
+++ b/docs/plans/2026-07-06-remix-v3-support-spike.md
@@ -122,17 +122,29 @@ It should not be the primary Remix v3 story.
 
 ## Open Work
 
-JSX-rich macros are the main unresolved part. The quick proof transforms after
-Remix's `node-tsx` has compiled JSX to `remix/ui/jsx-runtime` calls. That is
-fine for `t`, `msg`, `plural`, `select`, and `defineMessage`, but it is probably
-too late for `<Trans>`-style source JSX unless the native transform can support
-the lowered call shape.
+JSX-rich macros and browser-delivered modules are the main unresolved parts. The
+quick proof transforms after Remix's `node-tsx` has compiled JSX to
+`remix/ui/jsx-runtime` calls. That is fine for `t`, `msg`, `plural`, `select`,
+and `defineMessage`, but it is probably too late for `<Trans>`-style source JSX
+unless the native transform can support the lowered call shape.
+
+Reading `@remix-run/node-tsx@0.1.1` makes option 1 more concrete:
+`node-tsx` reads `.ts`, `.tsx`, and `.jsx` files itself and returns
+`shortCircuit: true`; it does not call `nextLoad()` for app files. Running
+Palamedes before Remix would therefore mean replacing that app-file pipeline:
+macro transform first, then the same OXC TS/JSX lowering behavior with
+tsconfig-aware options.
+
+Separately, browser-delivered modules go through Remix's asset compiler, not the
+Node `--import` hook chain. The register hook is therefore a server-executed
+module integration until Remix exposes a script transform hook for client
+assets.
 
 Likely follow-up choices:
 
 1. Run Palamedes before Remix's JSX transform, then hand the transformed source
-   to the same OXC TSX transform behavior. This may require owning more of the
-   loader pipeline.
+   to the same OXC TSX transform behavior. This requires owning the app-file
+   loader pipeline instead of composing with `node-tsx`.
 2. Teach the native transformer to recognize Remix-lowered JSX runtime calls for
    rich messages. This is more adapter-specific.
 3. Start Remix v3 support with JS macros plus server locale wiring, then add
--- a/docs/plans/2026-07-06-remix-v3-support-spike.md
+++ b/docs/plans/2026-07-06-remix-v3-support-spike.md
@@ -1,0 +1,161 @@
+# Spike: Remix v3 Support
+
+**Status:** Spike complete
+**Date:** 2026-07-06
+**Issue:** https://github.com/sebastian-software/palamedes/issues/284
+
+## Summary
+
+Remix v3 support looks feasible. The quickest useful integration path is a
+Node `--import` registration module that composes with Remix's own
+`remix/node-tsx` loader and runs Palamedes' existing native macro transform.
+
+The first proof covered server-side JS macros in a real Remix v3 beta scaffold.
+It did not yet solve the full UI adapter, PO catalog loading, or JSX-rich
+message components.
+
+## Current Remix v3 Shape
+
+Verified against `remix@next` on 2026-07-06:
+
+- `remix@next` resolved to `3.0.0-beta.5`.
+- `npx remix@next new` equivalent CLI shape is `remix new <target-dir>
+[--app-name <name>] [--force]`.
+- The generated app uses `node --import remix/node-tsx server.ts` for both
+  development and production scripts.
+- The scaffold requires Node `>=24.3.0`.
+- `remix/node-tsx` registers a synchronous `node:module` load hook via
+  `registerHooks({ load })`.
+- App modules are plain file imports under `app/`, with `server.ts`,
+  `app/router.ts`, `app/actions/controller.tsx`, and server rendering through
+  `remix/ui/server`.
+
+This matches the issue's core assumption: the default Remix v3 path is not a
+Vite pipeline, so the adapter should not start as a Vite-only integration.
+
+## Proof
+
+Scratch scaffold:
+
+```sh
+pnpm dlx remix@next new /private/tmp/palamedes-remix-v3-spike \
+  --app-name palamedes-remix-v3-spike \
+  --force
+```
+
+Built the minimum Palamedes transform chain in the spike worktree:
+
+```sh
+pnpm --filter @palamedes/transform... build
+```
+
+Added a temporary `palamedes-register.mjs` loader that:
+
+1. registers `node:module` `load`
+2. delegates to `nextLoad()` so Remix's `node-tsx` hook compiles TS/TSX
+3. runs `transformPalamedesMacros()` on returned JS source
+4. returns the transformed source to Node
+
+Then added a temporary Remix module:
+
+```ts
+import { t } from "@palamedes/core/macro"
+
+export function renderGreeting(name: string) {
+  return t`Hello ${name} from Remix 3`
+}
+```
+
+Smoke command:
+
+```sh
+node --import remix/node-tsx --import ./palamedes-register.mjs proof.mjs
+```
+
+Result:
+
+```text
+Hello Ada from Remix 3
+```
+
+Control command without the Palamedes hook:
+
+```sh
+node --import remix/node-tsx proof.mjs
+```
+
+Result: failed with the expected macro-stub error from
+`@palamedes/core/macro`, confirming the hook caused the transform.
+
+HTTP smoke:
+
+```sh
+PORT=44107 node --import remix/node-tsx --import ./palamedes-register.mjs server.ts
+curl -i http://127.0.0.1:44107/
+```
+
+Result:
+
+```text
+HTTP/1.1 200
+x-palamedes-proof: Hello Ada from Remix 3
+```
+
+## Recommended First Implementation
+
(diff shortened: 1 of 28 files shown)


packages/remix/src/index.ts
```diff
@@ -0,0 +1,82 @@
...
+    })
+
+    if (!result.hasChanged) {
+      return loaded
+    }
+
+    return {
+      ...loaded,
+      source: result.code,
+    }
```
swernerx, line 61: `result.map` gets dropped here, and that has a sneaky consequence: `remix/node-tsx` appends an inline `//# sourceMappingURL=` comment to the lowered source, and because our native transform patches via MagicString, that comment survives into the output — but the injected runtime import shifts every line after it. Node runs with `process.setSourceMapsEnabled(true)` in the Remix path, so stack traces in exactly the modules that contain macros will be mapped a line or more off. The fix doesn't need full map composition: strip the old inline comment and append `result.map` as a new inline sourcemap (it carries `sourcesContent`, so the mapping stays useful). Fine as a follow-up issue for an experimental package, but I'd track it — stale stack traces are the kind of papercut beta users notice first.

examples/remix-cookie/package.json
```diff
@@ -0,0 +1,21 @@
...
+    "build": "tsc --noEmit",
+    "dev": "PORT=4060 node --watch --import remix/node-tsx --import @palamedes/remix/register server.ts",
+    "start": "PORT=4060 node --import remix/node-tsx --import @palamedes/remix/register server.ts",
+    "typecheck": "tsc --noEmit"
+  },
+  "dependencies": {
+    "@palamedes/core": "workspace:^",
+    "@palamedes/remix": "workspace:^",
+    "@palamedes/runtime": "workspace:^",
+    "remix": "^3.0.0-beta.5"
```
swernerx: #284 explicitly called for pinning the example to an exact beta, and the rest of the matrix pins frameworks exactly (the react-router examples use `8.1.0`, no caret). `^3.0.0-beta.5` floats to beta.6+ on any fresh install, and with `@remix-run/node-tsx` at 0.1.1 the loader internals are precisely the part most likely to churn under us. I'd make this `3.0.0-beta.5`. The caret on the `@palamedes/remix` peer dependency is right, though — that one should stay a range.

examples/README.md
```diff
@@ -166,6 +166,12 @@ All matrix examples use the same public Palamedes stack:
 - `@palamedes/runtime`
 - `@palamedes/vite-plugin` or `@palamedes/next-plugin`
 
+Remix v3 support is being added through a separate beta example while Remix's
+component model settles:
+
+- [examples/remix-cookie](remix-cookie) - Node `--import` loader integration
```
swernerx, line 172: The manual curl smokes from the PR description map 1:1 onto a `smokeChecks` entry in `scripts/example-matrix.mjs` — port 4060, `accept-language: de`, assert on `Plätze frei`. The matrix entry shape doesn't assume the shared example UI, so this example fits as-is, and #284's acceptance criteria include wiring the example into the test setup. Doesn't need to happen in this PR, but a follow-up issue would keep it from getting lost.

examples/remix-cookie/app/i18n.ts
```diff
@@ -0,0 +1,71 @@
...
+export const locales = defineLocaleControls<Locale>({
+  locales: LOCALES,
+  defaultLocale: DEFAULT_LOCALE,
+  cookies: { locale: LOCALE_COOKIE },
+})
+
+export const LOCALE_LABELS = locales.labels
+export const normalizeLocale = locales.normalizeLocale
+
+const CATALOGS: Record<Locale, CatalogMessages> = {
```
swernerx, line 27: These hand-copied hash IDs are the fragile spot in the example: change one source string in `page.ts` and the hash moves, and the catalog silently falls back to the source language with no error anywhere. Every other example goes through `pmds extract` and `.po` files. If inline catalogs are deliberate to keep the first Remix example lean, a short comment saying so (and that the IDs come from the extractor) would stop the next person from tripping over it — otherwise I'd wire up the extract flow like the rest of the matrix.

packages/remix/README.md
```diff
@@ -0,0 +1,52 @@
+# @palamedes/remix
+
+Experimental Remix v3 integration for Palamedes.
+
+Use this package with Remix v3's default Node loader path. Register Remix's TSX
+loader first, then Palamedes:
+
+```sh
+node --import remix/node-tsx --import @palamedes/remix/register server.ts
```
swernerx, line 9: Two failure modes worth spelling out here, because both surface far from their cause:

1. The order of the two `--import` flags is load-bearing. Flip them and `remix/node-tsx` short-circuits without calling `nextLoad`, so our hook never sees `.ts`/`.tsx`/`.jsx` files — macros then blow up at request time with the generic "executed outside the compiler transform" error from `@palamedes/core/macro`. One sentence mapping that error to "check your `--import` order" will save real support round-trips.

2. This hook only reaches server-executed modules. Anything delivered to the browser is compiled on demand by `@remix-run/assets`, which has no script-transform hook — so client-side macros are out of scope for now. Saying that explicitly sets the right expectations for a beta integration.

swernerx (commented):
Did a deep pass on this, including reading the source of `@remix-run/node-tsx@0.1.1` and a fresh sweep of the Remix v3 beta repo/docs to sanity-check the integration model. Short version: the approach is right, and as far as I can tell it's the only one that composes with what Remix ships today.

Why I'm confident in the direction:

- Remix v3 has no plugin system and no Vite path. `--import` loader hooks are the sanctioned mechanism ("religiously runtime"), and the official scaffold runs the same loader chain in production (`start` = `node --import remix/node-tsx server.ts`, no build step). A register-hook package is the native shape for this, not a workaround.
- Node's hook chaining is LIFO, and node-tsx's load hook reads `.ts`/`.tsx`/`.jsx` from disk itself and returns `shortCircuit: true` without ever calling `nextLoad`. So `--import remix/node-tsx --import @palamedes/remix/register` is the only order that works at all, and transforming the post-lowered JS is the correct v1 tradeoff — details in the spike-doc comment.
- Nobody else is here yet: remix-i18next, inlang/paraglide and remix-intl all target Remix v2 / React Router. The first-mover claim in #284 holds.
- The AsyncLocalStorage request scope matches Remix's own model — they ship `@remix-run/async-context-middleware` on the same primitive.

The implementation reads clean: small surface, tests cover the hook contract and concurrent-request isolation, and the spike doc is genuinely useful. The inline comments cover what I'd fix or track: stale sourcemaps after the transform, exact-pinning the beta, wiring the example into the smoke matrix, the hand-copied catalogs, and two factual updates to README/spike doc. The most important of those: anything delivered to the browser goes through `@remix-run/assets`' own oxc compiler, which has no script-transform hook — so the register hook covers server-executed code only. That's the real boundary of "Remix v3 support" right now and worth stating out loud.

One follow-up idea beyond the comments: Remix ships `AcceptLanguage` in `remix/headers` plus first-party cookie middleware and typed request context in `fetch-router`. A small middleware in `@palamedes/remix` that negotiates the locale with those primitives and wraps `next()` in the i18n scope would turn the per-app wiring in the example into a one-liner — that's item 4 in #284.

docs/plans/2026-07-06-remix-v3-support-spike.md
```diff
@@ -0,0 +1,161 @@
...
+
+## Open Work
+
+JSX-rich macros are the main unresolved part. The quick proof transforms after
+Remix's `node-tsx` has compiled JSX to `remix/ui/jsx-runtime` calls. That is
+fine for `t`, `msg`, `plural`, `select`, and `defineMessage`, but it is probably
+too late for `<Trans>`-style source JSX unless the native transform can support
+the lowered call shape.
+
+Likely follow-up choices:
```
swernerx, line 143: Having read `@remix-run/node-tsx@0.1.1`'s source: option 1 is harder than "may require owning more of the loader pipeline" suggests, and at the same time cheaper than it sounds. node-tsx's load hook reads files from disk itself and returns `shortCircuit: true` — it never calls `nextLoad` for `.ts`/`.tsx`/`.jsx` — so there is no way to feed it pre-transformed source. Running Palamedes first means fully replacing node-tsx for app files: macro pass plus our own JSX/TS lowering with the same tsconfig-driven options. The good news is that node-tsx is just `oxc-transform` + `get-tsconfig` + an inline sourcemap, and we already sit on OXC, so owning that pipeline is a small amount of code.

Also worth adding to Open Work: the client path. Browser-delivered modules go through `@remix-run/assets` (its own oxc compile, no transform hook), so the register hook covers server-executed code only. That, more than JSX lowering, is the real boundary for the "first i18n library with Remix v3 support" claim.

Issue #357, opened 2026-07-06
Epic: complete full-stack Remix v3 support


swernerx:
## Outcome

Make `@palamedes/remix` a trustworthy full-stack Remix v3 integration: ordinary and rich Palamedes macros compile in server and browser module graphs, Remix UI can render compiled rich messages, the browser receives the correct locale/catalog, and the complete path is covered by a representative example and automated browser tests.

This epic continues the work started in #284 and the shipped server path from #353.

## What changed

The original follow-up was blocked on a public Remix browser asset transform API. Remix now documents `createAssetServer({ scripts: { loaders } })`; these loaders use the synchronous Node load-hook contract and run after Remix's TypeScript/JSX compilation:

- https://github.com/remix-run/remix/blob/main/packages/assets/README.md#loaders
- upstream request: https://github.com/remix-run/remix/issues/11580

A focused beta.10 probe established:

- the existing Palamedes load hook can transform ordinary macros in browser-delivered modules;
- Remix-lowered `_jsx`/`_jsxs` calls retain the structure needed to reconstruct rich messages;
- the current transformer does not yet recognize that lowered representation;
- direct browser `.po` imports fail before script loaders run, so client catalog delivery needs a separate contract.

The upstream hook blocker is therefore resolved. The remaining work is Palamedes-side compilation, runtime, bootstrap, catalog, and quality work.

### Now — establish the full-stack runtime path

- [ ] #1026 — integrate the Palamedes transform with Remix browser assets
- [ ] #1029 — compile rich macros from Remix-lowered JSX runtime calls
- [ ] #1027 — implement the Remix UI compiled rich-message runtime
- [ ] #1031 — bootstrap client i18n and deliver catalogs without browser `.po` imports

The transform, UI runtime, and bootstrap work may proceed in parallel once their shared compiled-message contract is agreed.

### Next — make the path trustworthy

- [ ] #1028 — preserve source maps, HMR, and config/catalog invalidation
- [ ] #1032 — prove SSR, browser execution, rich messages, and locale navigation end to end

### Release/readiness

- [ ] #1030 — update support claims, setup documentation, version policy, and demo-readiness criteria

## Quality bar

The epic is complete when:

- ordinary and rich Palamedes macros work in Remix server and browser module graphs;
- authored TSX extraction and post-Remix transformation agree on message identity and placeholders;
- SSR and browser rendering use the same locale/catalog without divergence;
[...]

Issue #353, opened 2026-07-06
Bring Remix v3 support to server-first Palamedes parity


swernerx:
## Context

PR #350 merged the first experimental Remix v3 integration:

- `@palamedes/remix` with a Node `registerHooks()` loader for server-loaded JS/TS modules
- request-local i18n scope for Remix handlers
- `examples/remix-cookie` with cookie locale switching, `Accept-Language` negotiation, catalog loading, and matrix smoke coverage

That is a strong server-side candidate. This issue tracks the next step: bring Remix v3 to server-first Palamedes parity where Remix's beta architecture allows it, while making the remaining UI/client boundaries explicit.

Refs #284
Follow-up to #350

## Goal

Make Remix v3 a first-class server-side Palamedes integration:

- normal catalog files and extraction workflow
- reusable Remix server API instead of per-example wiring
- standard Palamedes locale strategy examples
- clear documentation for supported macros, unsupported UI/client paths, and Remix beta tracking

Full client-side feature parity is not yet a realistic promise because Remix v3 browser-delivered modules currently go through Remix's asset compiler, not the Node `registerHooks()` chain. This issue should make that boundary visible and start the upstream/API work needed to remove it.

## Spikes Before Implementation

Two risks are cheap to check up front and would change the design if they fail. Run these before the corresponding slice.

### Spike 1: `.po` imports through the real Remix loader chain (gates Slice A)

Claiming `.po` URLs in the load hook is expected to work — Node's resolver is extension-agnostic, and the default loader never sees the file because the hook returns the module itself. But it has not been proven inside a real `remix/node-tsx` chain. Half a day, throwaway scaffold:

- [x] `import { messages } from "./locales/de.po"` resolves and executes under `node --import remix/node-tsx --import @palamedes/remix/register`.
- [ ] `node --watch` (the dev script) picks up `.po` edits, i.e. catalog files enter the watched module graph.
- [x] `tsc --noEmit` passes with the `po.d.ts` declaration pattern the other examples use.

### Spike 2: request scope vs. streamed rendering (gates the Slice B API shape)

The middleware plan wraps `next()` in `scope.run()`. That covers everything that executes before the `Response` is returned. If `remix/ui/server` renders lazily into the response body stream — and Frames fragments likely use the same path — translated code may run while the body is being consumed, after `run()` already resolved and potentially outside the AsyncLocalStorage context. If that context is lost, `getI18n()` silently resolves nothing and the middleware API has to look different.

[...]

PR #359, opened 2026-07-06
Bring Remix v3 to server-first parity

swernerx:
## Summary

Refs #353.

This brings the Remix v3 integration to the server-first parity path described in the epic:

- loads checked-in `.po` catalogs through `@palamedes/remix/register` and adds sync config loading for the Node loader path
- adds `@palamedes/remix/server` helpers for locale resolution, request-local i18n, catalog message caching, cookie serialization, and fetch-router middleware
- expands Remix examples to cookie, route, subdomain, and TLD strategies with checked-in catalogs and smoke coverage
- documents the supported server-first setup, tested Remix beta, unsupported browser/client and rich JSX paths, migration notes, and beta tracking
- adds a non-blocking `remix@next` canary smoke job

## Follow-ups

- Remix asset pipeline transform hook: https://github.com/remix-run/remix/issues/11580
- Remix UI adapter, rich messages, and Frames: #357

## Verification

- `pnpm check:release-set`
- `pnpm format:check`
- `pnpm lint:oxlint`
- `pnpm build`
- `pnpm check-types`
- `pnpm test`
- `pnpm verify:examples:smoke -- --framework remix`


packages/remix/src/server.ts
```diff
@@ -31,3 +94,168 @@ export function createRemixI18nRequestScope<T extends I18nInstance = I18nInstanc
...
+    },
+
+    get(context) {
+      if (context) {
+        return context.get(remixI18nContext) as RemixI18nContextValue<TLocale, T> | undefined
+      }
+
+      const i18n = scope.get()
+      return i18n
+        ? ({ i18n, locale: i18n.locale as TLocale, source: "default" } as const)
```
swernerx: When `get()` is called without a router context, `source` always comes back as `"default"` even if the request locale was actually resolved from a cookie or the route — the AsyncLocalStorage scope only carries the i18n instance, so the real resolution source is lost here. Anything downstream that branches on `source` would get misleading data. A cheap fix: populate a module-level `WeakMap<i18n, RemixI18nContextValue>` in `run()` and look the full context value up from it here. If that's not worth it yet, a doc note that scope-based `get()` doesn't preserve `source` would do.

packages/remix/src/server.ts
```diff
@@ -31,3 +94,168 @@ export function createRemixI18nRequestScope<T extends I18nInstance = I18nInstanc
...
+          {
+            headers: context.headers,
+            params: context.params,
+            request: context.request,
+          },
+          async (palamedes) => {
+            context.set(remixI18nContext, palamedes, { property: "palamedes" })
+            return await next()
+          }
+        )) as RemixI18nServer<TLocale, T>["middleware"] extends () => infer TMiddleware
```
swernerx: This self-referential conditional-type cast is hard to follow — the type it extracts is the `Middleware<...>` shape already spelled out in `RemixI18nServer`. Typing the value directly keeps the compiler honest and reads better:

```ts
const middleware: ReturnType<RemixI18nServer<TLocale, T>["middleware"]> = async (context, next) => ...
return middleware
```

Taste thing, take it or leave it.

swernerx (commented):
Read through this against #353 — Slice A and B land cleanly, the streamed-body scope binding answers Spike 2 with a real test, and the support matrix is honest about the client boundary. A few notes on the lines below; none of them block.

examples/remix-route/app/i18n.ts
```diff
@@ -0,0 +1,78 @@
...
+export const locales = defineLocaleControls<Locale>({
+  locales: LOCALES,
+  defaultLocale: DEFAULT_LOCALE,
+  hosts: {
+    locales: {
+      en: "en.lvh.me",
+      de: "de.lvh.me",
+      es: "es.lvh.me",
+    },
+  },
```
swernerx: Is this host mapping intentional in the route example? The strategy here is `route`, and `suggest()` falls back to accept-language just fine without host config — but with this mapping in place, a request arriving via `en.lvh.me` while rendering `/de` gets a host-based switch suggestion, which is confusing behavior for a route-strategy demo. It looks like it traveled over from the subdomain setup. While in here: `LOCALE_COOKIE` is exported but unused in this example.

.github/workflows/ci.yml
```diff
@@ -111,3 +111,42 @@ jobs:
...
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Override Remix examples to remix@next
+        run: pnpm --filter "@palamedes/example-remix-*" add remix@next --save-exact
+
+      - name: Smoke-test Remix examples against remix@next
+        id: remix-next-smoke
+        continue-on-error: true
```
swernerx, line 142: With `continue-on-error` the job stays green when the canary fails, and the step summary is only visible after opening the run. An annotation in the failure step (`echo "::warning::remix@next canary failed — see step summary"`) would surface the signal in the checks UI without clicking through. Optional.

packages/remix/src/server.ts
swernerx: Fixed in `baf7478`: `createRemixI18nServer()` now keeps the full resolved context in a `WeakMap` keyed by the active i18n instance, so scope-based `get()` preserves `source` instead of rebuilding it as `default`. Added coverage for both `accept-language` and `cookie` sources.

packages/remix/src/server.ts
swernerx: Taken in `baf7478`: the middleware value is now typed directly as `ReturnType<RemixI18nServer<TLocale, T>["middleware"]>` and returned explicitly, which removes the self-referential conditional cast.

examples/remix-route/app/i18n.ts
swernerx: Fixed in `baf7478`: the route example no longer carries the subdomain host mapping, the unused `LOCALE_COOKIE` export is gone, and the route banner suggestion no longer passes cookie/host inputs.

.github/workflows/ci.yml
swernerx: Added in `baf7478`: the canary failure summary now also emits a `::warning::` annotation so the non-blocking signal is visible from the checks UI.

Issue #355, opened 2026-07-06
Extractor drops plural messages depending on sub-message content (empty msgid)


swernerx:
## Summary

`pmds extract` (release CLI, `@palamedes/core/macro` `plural()`) **silently drops plural messages depending on the plural sub-message content**, writing an empty `msgid ""` instead of the ICU plural msgid. Lingui and i18next-parser extract the same logical messages correctly, so this is specific to the Palamedes extractor.

Discovered while building a realistic benchmark corpus in `benchmarks/e2e-workflow` (mixing plain / variable / plural messages). Of 210 synthetic plurals, only **49 extract**; the other 161 produce empty msgids. The set that extracts is scattered/unpredictable across otherwise-identical message shapes, so it depends on the concrete sub-message text.

## Reproduction

`palamedes.yaml` with a `src/locales/{locale}` catalog including `src/generated`, and:

```ts
import { plural, t } from "@palamedes/core/macro"

export function Demo(count) {
  // WORKS -> msgid "{count, plural, one {# queue detail 003} other {# queue details 003}}"
  const a = plural(count, { one: "# queue detail 003", other: "# queue details 003" })
  // FAILS -> msgid ""
  const b = plural(count, { one: "# queue detail 00042-now", other: "# queue details 00042-now" })
  return [t({ message: "x" }), a, b]
}
```

## What was ruled out

Bisecting single factors — none of these are the cause:

- `count` as a free identifier vs a function parameter vs `const count = 1`
- returning a bare array vs `.join("\n")`
- one plural vs several plurals in a file
- `.ts` vs `.tsx` extension

The only factor that flips extraction success/failure is the **sub-message text itself**. Short numeric tokens like `003`, `42` extract; longer/alphanumeric ones like `00003`, `12345`, `00000`, `213-now`, `n00042`, `00042n`, `0042-x`, `item00042` fail. Within a padded `000`..`209` range the successes are scattered (e.g. `002 007 009 014 019 021 ...`), so it is not a simple length rule.

## Impact

Message content should never determine whether extraction succeeds. This makes `plural()` unreliable and blocks a realistic (plural-containing) benchmark corpus for the extract/update workflow benchmark.

## Suggested next steps

- Investigate the plural sub-message parse/serialize path in the Rust core / ferrocat ICU handling.
- Add a regression test covering plural sub-messages that contain multi-digit numbers, hyphens, and trailing words.

PR #360, opened 2026-07-06
Homepage polish + realistic extract-time benchmark

swernerx:
Two related pieces of work on the marketing site and the e2e-workflow benchmark.

## 1. Homepage polish (`065946d`)

A review pass over the new homepage layout:

- **Hero:** add the large Palamedes logo medallion above the headline.
- **Tab highlight fix:** Base UI (RC) emits `aria-selected`, not `data-selected`, so the `data-[selected]` utilities on the Write/Extract/Translate tabs never matched — the active tab was indistinguishable. Switched to `aria-selected` + a bronze top rule.
- **TerminalCascade:** stop the size jump (report bars are now always reserved and fade in) and give the three panels equal width with a clean gap (was a negative-margin overlap plus a right indent that made the lower two narrower).
- **Nav:** wider header link padding for proper rhythm.
- **WorkflowFlow:** new Write → Extract → Translate → Render diagram above the code showcase.
- **LocaleBookingCards:** render the three locales across the full width instead of a narrow left column.
- **Section 04 (Scope):** rewritten from architecture jargon (`transform`, `ferrocat`, "locale negotiation") into plain developer language, with a scope diagram contrasting what Palamedes owns (same on every framework) vs what your framework owns (different per stack), and the payoff line "switch frameworks, and only the bottom row changes."
- **BenchmarkChart:** drop the in-bar ratio annotation, round timings to whole ms, align axis labels under the bars.
- **Matrix:** subdomain demos are live (verified reachable — all 15 return 200 and render their locale) so they're promoted from `provisioning` to live links; TLD stays `provisioning` (domains still unresolved). Stale reachability note in `docs/demo-deployments.md` corrected.

## 2. Realistic extract-time benchmark (`d02d505`)

The e2e-workflow `realistic` corpus was an all-i18n fixture — every file dense with messages. Real apps aren't like that: the extractor spends most of its time scanning ordinary, non-i18n source. Rebuilt to model a production web app's Lingui include roots (figures rounded so they read as a shape, not false precision):

- **1,500 files**, of which **750 carry i18n markers** and 750 have none but still get scanned.
- **~400,000 source lines**, ~3% of them i18n; files padded with deterministic non-i18n filler to the line budgets.
- **6,000 messages** (plain + ~15% `{name}` variables) spread thinly across the marked files.

This makes the timed workload reflect real parse volume rather than a best-case dense fixture. Fresh numbers: **Palamedes 174 ms vs Lingui 2254 ms (12.99×)** and i18next-parser 1562 ms (9.00×). Home, proof, compare, prose and the methodology docs quote the new numbers; the drift guard (`verify-site-bench-data`) stays green. The chart gained an adaptive tick step so a larger range stays at ~8-10 gridlines.

### Note on the PO reader / #355 / #358

While building the corpus I mis-diagnosed a "dropped messages" symptom as an extractor bug (#355). It was actually the benchmark's PO reader: the native extractor emits valid **multiline** PO msgids for longer messages, and the old single-line reader treated them as empty. This PR includes the multiline-reader fix (`benchmarks/e2e-workflow/src/po.mjs`), identical to the now-merged **#358** — it should reconcile cleanly on merge. The native extractor was correct.

## Validation

- `pnpm --dir site typecheck` — clean
- `node scripts/verify-site-bench-data.mjs` — bench.ts matches the checked-in report
- `oxlint` on changed files — clean
- Benchmark re-run; all three tools extract the same 6,000 messages

### DIFF
--- a/benchmarks/e2e-workflow/results/latest.json
+++ b/benchmarks/e2e-workflow/results/latest.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "benchmark": "palamedes-e2e-extract-update-workflow",
-  "generatedAt": "2026-07-06T09:09:36.849Z",
+  "generatedAt": "2026-07-06T18:05:29.525Z",
   "machineLocal": true,
   "seed": 20260703,
   "warmup": 3,
@@ -11,12 +11,12 @@
     "nodeVersion": "v24.18.0",
     "platform": "darwin",
     "arch": "arm64",
-    "generatedAt": "2026-07-06T09:09:36.849Z"
+    "generatedAt": "2026-07-06T18:05:29.525Z"
   },
   "versions": {
     "benchmarkPackage": "@palamedes/benchmark-e2e-workflow",
     "palamedes": {
-      "cli": "pmds (Palamedes) v1.2.0"
+      "cli": "pmds (Palamedes) v1.3.0"
     },
     "lingui": {
       "cli": "6.4.0"
@@ -70,7 +70,7 @@
       "results": [
         {
           "tool": "palamedes",
-          "version": "pmds (Palamedes) v1.2.0",
+          "version": "pmds (Palamedes) v1.3.0",
           "profile": "small",
           "fileCount": 80,
           "sourceMessageCount": 640,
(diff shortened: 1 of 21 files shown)


swernerx:
Pushed a fix + addressed the review feedback:

- **CI (formatting):** the `validate` job failed only on `pnpm format:check` — `benchmarks/e2e-workflow/src/corpus.mjs` needed oxfmt line-wrapping. Fixed; `format:check`, `lint` and the drift guard are green locally. (The other CI steps hadn't run — the job stops at the first failure.)
- **Greptile P2 — `ratios` field:** not dead. `scripts/verify-site-bench-data.mjs` reads it and asserts the speedups against the checked-in report, so the ratios quoted in prose (hero, ProofStrip) can't silently drift, even though the chart no longer renders them inline. Added a comment on the field saying so.
- **Greptile P2 — `BENCH_SMALL` / `BENCH_MEDIUM`:** also not unused. They're read by the same drift guard (`parseBenchSection("SMALL"/"MEDIUM")`) and back the tables in the benchmark docs; only `BENCH_REALISTIC` is charted on the site. Added a comment documenting that intent.

Kept both rather than removing — they're the checked-in reference for the smaller corpora and are guarded against drift.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #429, opened 2026-07-24
fix(core): support ICU apostrophe quoting

### LINKED ISSUE #408: fix(core): no ICU apostrophe-quoting support — migrated Lingui/TMS catalogs corrupt silently, literal { is unrepresentable
## Summary

The runtime message parser has no quote handling anywhere (`packages/core/src/messageFormat.ts:71-121`). Verified by execution:

- `It''s done` → `It''s done` (ICU: `It's`)
- `Literal '{name}' here` → `Literal 'X' here` (ICU: literal `{name}`)
- `{n, plural, other {'#' of #}}` → `'5' of 5` (ICU: `# of 5`)
- A bare literal `{` (e.g. `Use { to open`) always **throws** — there is no escape mechanism at all for literal `{`, `}`, `#`. This feeds the `<Trans>` crash issue.

The Rust toolchain deliberately adopts `IcuSyntaxPolicy::RuntimeLiteralApostrophes` (translators needn't double apostrophes — a reasonable choice), but the flip side is undocumented: Lingui catalogs and TMS exports legitimately containing ICU quoting (`''` doubling is standard vendor output) silently corrupt on migration.

## Suggested fix

Either implement the ICU quoting subset (`''` → `'`, `'{...}'` literal, `'#'` literal inside plural), or:

1. document the divergence prominently in `docs/migrate-from-lingui.md`, and
2. make `pmds audit` flag `''` / `'{` sequences in imported catalogs so migrations surface the incompatibility instead of shipping corrupted strings.

swernerx:
## What changed

- implement ICU MessageFormat apostrophe quoting in the core runtime parser
- preserve ordinary apostrophes while decoding doubled apostrophes
- represent quoted braces and plural `#` syntax as literal message nodes
- render literal nodes consistently in the string, React, and Solid consumers
- add regression coverage for the reported migration and TMS patterns

## Why

The runtime parser treated every brace and plural pound sign as syntax and left doubled apostrophes untouched. Valid ICU/Lingui/TMS patterns could therefore render corrupted text or throw when they contained literal syntax characters.

The parser now follows ICU's apostrophe-friendly “only where needed” behavior: an apostrophe starts quoting only before `{`, `}`, or `#` in plural text, while `''` always represents one literal apostrophe.

## Impact

Migrated ICU catalogs can represent literal braces and plural pound signs without interpolation, and standard doubled-apostrophe output renders correctly across the core, React, and Solid runtimes.

## Validation

- `pnpm format:check`
- `pnpm lint`
- `pnpm check:release-set`
- `pnpm proof:icu:check`

Fixes #408

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -152,6 +152,7 @@ export type {
   MessageNode,
   MessageChoiceNode,
   MessageFormattedArgumentNode,
+  MessageLiteralNode,
   MessageTagNode,
   MessageVariableNode,
 } from "./messageFormat"
--- a/packages/core/src/messageFormat.test.ts
+++ b/packages/core/src/messageFormat.test.ts
@@ -51,6 +51,15 @@ describe("formatMessagePattern", () => {
     expect(formatMessagePattern("This '{isn''t}' obvious")).toBe("This {isn't} obvious")
   })
 
+  it("auto-closes an unterminated quoted sequence at the end of the pattern", () => {
+    expect(
+      formatMessagePattern("Send '{name} to {target}", {
+        name: "Alice",
+        target: "Bob",
+      })
+    ).toBe("Send {name} to {target}")
+  })
+
   it("keeps quoted arguments and plural pound signs literal", () => {
     expect(formatMessagePattern("Literal '{name}' here", { name: "value" })).toBe(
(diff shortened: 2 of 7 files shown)


swernerx:
@greptileai

PR #430, opened 2026-07-24
fix(ui): make Trans catalog parsing resilient

### LINKED ISSUE #403: fix(react): &lt;Trans&gt; throws on malformed catalog patterns instead of falling back to the source message
## Summary

`Trans` calls `i18n.getMessage(...)` (raw pattern) and then `parseMessagePattern(pattern)` directly (`packages/react/src/index.tsx:55-64`) — completely outside `createI18n`'s try/catch + `onError` + source-fallback machinery (`packages/core/src/index.ts:90-120`).

The core README explicitly promises: "Malformed runtime patterns fall back to the source message instead of throwing through the component tree" (`packages/core/README.md:33-37`). The `<Trans>` path violates that guarantee.

## Reproduction (verified by execution)

`parseMessagePattern("{count, plural one {...}}")` — a missing comma, i.e. a one-character translator typo in a PO entry — throws `Expected "," while parsing message pattern.`

In a production app loading catalogs at runtime, one bad PO entry makes every page containing that `<Trans>` crash: a 500 on the server, an error boundary on the client. `onError` never fires.

## Suggested fix

Wrap parsing in `Trans` in try/catch, fall back to `metadata.message ?? id`, and route the failure through the instance's `onError`. That likely means exposing a resilient nodes-oriented API from core (e.g. `i18n.getMessageNodes(id, metadata)`) rather than having the component consume the raw-pattern `getMessage`.

Related: since the runtime parser has no escape mechanism for a literal `{` (see the ICU apostrophe-quoting issue), innocuous-looking catalog content can currently reach this throwing path.

swernerx:
## Summary
- add a resilient `getMessageNodes` path to the core i18n instance
- route React and Solid `Trans` parsing through the same `onError` and source-fallback behavior as string formatting
- render malformed source patterns as plain text and cover catalog/source failures with core and integration tests

## Verification
- `pnpm format:check`
- `pnpm lint`
- `pnpm check:release-set`

Fixes #403

### DIFF
--- a/packages/core/src/index.test.ts
+++ b/packages/core/src/index.test.ts
@@ -155,6 +155,32 @@ describe("createI18n", () => {
     expect(errors).toStrictEqual(["{count, plural one {# file} other {# files}}"])
   })
 
+  it("returns fallback nodes when a catalog pattern is malformed", () => {
+    const errors: string[] = []
+    const i18n = createI18n({
+      onError(info) {
+        errors.push(info.pattern)
+      },
+    })
+
+    i18n.load("de", {
+      greeting: "Hallo {name",
+    })
+    i18n.activate("de")
+
+    expect(i18n.getMessageNodes("greeting", { message: "Hello {name}" })).toStrictEqual([
+      { type: "text", value: "Hello " },
+      { type: "variable", name: "name" },
+    ])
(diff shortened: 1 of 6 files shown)


swernerx:
@greptileai

PR #431, opened 2026-07-24
fix(cli): fail safely on extraction errors

### LINKED ISSUE #405: fix(cli): extraction failures are invisible without --verbose, and failed files' messages get marked obsolete
## Summary

`failed_files` warnings are printed only when `--verbose` is set (`crates/palamedes-cli/src/main.rs:386-393`), while the summary line prints `✓ Extracted N messages from M files` unconditionally (`main.rs:337`).

## Failure scenario

Any file that fails to parse or hits a non-fatal extraction error (e.g. `UnnamedPlaceholder` — see the placeholder-name divergence issue — or a syntax edge oxc rejects) is silently dropped. Because the aggregated message list no longer contains that file's messages, `update_catalog_file` marks all of them obsolete in **every** locale (`ObsoleteStrategy::Mark`, `catalog_update.rs:256-262`), and `--clean` deletes them after the 30-day window (`catalog_update.rs:376-381`).

A CI pipeline running plain `pmds extract` reports success while the catalogs silently decay — translations get orphaned and eventually deleted without anyone noticing.

## Suggested fix

- Always print failed-file warnings to stderr (not gated on `--verbose`).
- Include a failure count in the summary line.
- Exit non-zero when `failed_files` is non-empty (or add a `--strict` flag for CI use).

swernerx:
## Summary
- report every skipped source-file extraction failure on stderr regardless of `--verbose`
- collect all catalog extraction results before writing any locale catalog
- fail with a counted non-zero error when any source file failed, preventing messages from being marked obsolete or cleaned
- add a regression test proving existing catalogs remain unchanged and new locale catalogs are not created after a parse failure

## Verification
- `RUSTUP_TOOLCHAIN=1.95 cargo fmt --all -- --check`

Fixes #405

### DIFF
--- a/crates/palamedes-cli/src/main.rs
+++ b/crates/palamedes-cli/src/main.rs
@@ -214,6 +214,8 @@ enum CliError {
     AuditFailedOnWarning { errors: usize, warnings: usize },
     #[error("Catalog completeness below {threshold} for {locales}.")]
     CompletenessBelowThreshold { threshold: String, locales: String },
+    #[error("Extraction failed for {failures} source file(s); catalogs were not updated.")]
+    ExtractionFailed { failures: usize },
     #[error("Could not build glob pattern {pattern}: {source}")]
     GlobPattern {
         pattern: String,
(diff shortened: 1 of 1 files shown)


swernerx:
@greptileai Please re-review the latest commit, including the watch-mode recovery regression test.

PR #434, opened 2026-07-24
fix(runtime): isolate SSR client locale activation

### LINKED ISSUE #402: fix(react/runtime): SSR of client components mutates a shared i18n singleton — cross-request locale bleed
## Summary

The documented client bridge `useClientLocale(locale, sync)` calls `sync(locale)` unconditionally during render — including the SSR pass of client components (`packages/react/src/client.tsx:1-6`). The wiring all four Next.js examples prescribe (`examples/nextjs-*/src/lib/i18n.client.ts:19-36`) makes `sync` do two things on the server:

1. `clientI18n.load(locale)` / `activate(locale)` on a **module-level singleton** shared by all concurrent requests, and
2. `setServerI18nGetter(() => clientI18n)`, permanently replacing the getter in the SSR module graph with one that ignores the AsyncLocalStorage request scope set up by `createServerI18nScope`. `serverI18nGetter` is a plain module-level `let` (`packages/runtime/src/index.ts:10-11,51-53`), not request-scoped.

## Failure scenario

Under streaming SSR, request A (`de`) activates the singleton and suspends; request B (`en`) activates it; A's remaining client components render in English → mixed-language HTML within one response, plus hydration mismatches when the client re-syncs. Single-user dev traffic never interleaves, so this ships unnoticed and only appears under production load.

Render-phase mutation of shared state also violates React's concurrent-rendering rules: a thrown-away/replayed render (StrictMode, Suspense, transitions) still commits the global locale switch.

## Suggested fix

- On the server, never route SSR of client components through the shared singleton. Feed a request-scoped instance (the ALS mechanism in `packages/runtime/src/server.ts` is sound and already global across bundling layers via `Symbol.for`) — e.g. a provider that creates/activates a per-request instance, keeping the ALS getter authoritative.
- On the client, drive updates through `useSyncExternalStore` over `subscribeClientI18n` instead of render-phase `sync()`.
- Consider making `@palamedes/react` fail loudly if a locale-activating sync is invoked during server render.
- Update all `examples/nextjs-*` (and other frameworks with the same wiring) accordingly — the plugin README even warns against per-render `setServerI18nGetter()` calls, which is effectively what the examples do.

Found during a pre-1.5.1 production-readiness review (verified against source by two independent passes).

swernerx:
## Summary

- prevent `useClientLocale` from invoking client synchronization during server rendering
- bridge React locale updates through `useSyncExternalStore`, including re-activation of the same mutable i18n instance
- connect Turbopack's isolated client-component SSR runtime to the existing request-local `AsyncLocalStorage` scope
- update all four Next.js examples to use fresh request-local SSR instances and pre-hydration client initialization
- keep server component and server action translations inside explicit request scopes

## Verification

- `CI=true pnpm format:check`
- `CI=true pnpm lint`
- `RUSTUP_TOOLCHAIN=1.95 cargo fmt --all -- --check`
- production builds for the cookie, route, subdomain, and TLD Next.js examples
- `CI=true RUSTUP_TOOLCHAIN=1.95 node scripts/verify-examples-browser.mjs --framework nextjs`

Fixes #402

### DIFF
--- a/docs/api/react.md
+++ b/docs/api/react.md
@@ -66,7 +66,7 @@ After the client commits, it synchronizes the locale and subscribes to
 `setClientI18n()` activations so translated descendants re-render.
 
 Initialize the client i18n before hydration when translated client components
-render in the initial HTML. The hook retains a synchronous browser-only
-bootstrap for existing integrations, but pre-hydration initialization avoids
-render-phase state changes. The sync function may return a promise; the hook
-intentionally does not own loading UI or routing policy.
+render in the initial HTML. The hook never invokes the sync callback during
+render; this avoids external-store mutations during both SSR and client render
+retries. The sync function may return a promise; the hook intentionally does
+not own loading UI or routing policy.
--- a/docs/api/react.md
+++ b/docs/api/react.md
@@ -61,6 +61,12 @@ const items = buildLocaleSwitchItems({
 })
 ```
 
-`useClientLocale(locale, sync)` calls `sync(locale)` from a client component.
-The sync function may return a promise; the hook intentionally does not own
-loading UI or routing policy.
+`useClientLocale(locale, sync)` never calls `sync` during server rendering.
+After the client commits, it synchronizes the locale and subscribes to
+`setClientI18n()` activations so translated descendants re-render.
+
+Initialize the client i18n before hydration when translated client components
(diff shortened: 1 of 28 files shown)


swernerx:
@greptileai Please review the current head commit. The initial Greptile check has remained pending for over 20 minutes without publishing a review.

PR #436, opened 2026-07-24
fix(i18n): support plural offsets end to end

### LINKED ISSUE #407: fix: plural offset is broken across the whole pipeline — runtime mis-renders it, extractor drops or corrupts it
## Summary

ICU plural `offset:` is unsupported end-to-end, but every stage fails **silently**, so a Lingui migrator using `plural(count, { offset: 1, ... })` gets wrong text in production with no diagnostic.

## Runtime (verified by execution)

The runtime parser has no concept of `offset:` (`packages/core/src/messageFormat.ts:155-166` option parsing, `:462-483` selection). For `{n, plural, offset:1 =0 {nobody} one {you and one other} other {you and # others}}` the option loop stores a bogus key `"offset:1 one"`. With `n = 2` the output is `"you and 2 others"` instead of ICU-correct `"you and one other"` — the `one` branch is unreachable and `#` renders `n` instead of `n - offset`. Meanwhile the compiler transform emits `offset` into runtime patterns (`crates/palamedes/src/transform/messages.rs:580-591`, `transform/runtime.rs:428`) and the audit/compile toolchain (ferrocat-icu) parses `offset:` as valid ICU — so `pmds audit` passes catalogs the runtime then renders wrongly. Mirror handling is also needed in `@palamedes/react`'s `selectChoice`/pound replacement.

## Extract/transform (verified empirically)

- `<Plural value={count} offset={1} .../>` (numeric — the Lingui-idiomatic form) → `offset:1` is silently dropped on both extract and transform sides (`crates/palamedes/src/extract.rs:998-1035` keeps only string-valued attributes). Only the string form `offset="1"` works.
- `plural(count, { offset: 1, ... })` → silently dropped (`extract.rs:563-599`, call form always passes `offset: None` at `:803`).
- `plural(count, { offset: "1", ... })` → **corrupted msgid** `{count, plural, offset {1} one {# item} ...}` (offset treated as a category); `ferrocat::derive_message_metadata_from_icu` accepts it without a diagnostic, so neither audit nor compile flags it.

## Suggested fix

1. Runtime: parse `offset:N` in `parseBraceExpression` (store on `MessageChoiceNode`), subtract it before `Intl.PluralRules.select` and in `#` replacement; exact `=N` keys match the un-offset value per ICU.
2. Extract/transform: handle the `offset` key explicitly in both choice extractors (accept numeric literal and string) and reject unknown non-category keys instead of emitting them into the ICU message.
3. If a full runtime fix is deferred, make the compiler/audit **reject** `offset` as unsupported instead of silently rendering it wrong.

swernerx:
## Summary

- parse ICU plural offsets into the runtime AST and apply them to category selection and `#` replacement while keeping exact matches on the original value
- preserve static numeric and string offsets in call and JSX macro extraction/transformation
- reject dynamic offsets and invalid plural categories instead of silently dropping or corrupting them
- keep direct and rich-message rendering aligned across Core, React, and Solid

## Root cause

The compiler could emit `offset:N`, but the runtime parsed it as part of an option key and selected plural branches using the unadjusted value. Separately, the macro pipeline ignored numeric offsets and treated string offsets as ordinary categories in call expressions.

## Validation

- `CI=true pnpm format:check`
- `CI=true pnpm lint`
- `CI=true pnpm check:published-types`

Closes #407

### DIFF
--- a/crates/palamedes/src/choice.rs
+++ b/crates/palamedes/src/choice.rs
@@ -1,0 +1,70 @@
+const PLURAL_CATEGORIES: [&str; 6] = ["zero", "one", "two", "few", "many", "other"];
+const MAX_SAFE_INTEGER: f64 = 9_007_199_254_740_991.0;
+
+pub(crate) fn is_plural_format(format: &str) -> bool {
+    matches!(format, "plural" | "selectordinal")
+}
+
+pub(crate) fn normalize_choice_option_key(format: &str, key: &str) -> Option<String> {
+    if !is_plural_format(format) {
+        return Some(key.to_string());
+    }
+
+    if PLURAL_CATEGORIES.contains(&key) {
+        return Some(key.to_string());
+    }
+
+    let exact = key.strip_prefix('_').or_else(|| key.strip_prefix('='))?;
+    is_non_negative_integer(exact).then(|| format!("={exact}"))
+}
+
+pub(crate) fn normalize_string_offset(value: &str) -> Option<String> {
+    let value = value.trim();
+    let parsed = value.parse::<u64>().ok()?;
+    ((parsed as f64) <= MAX_SAFE_INTEGER).then(|| value.to_string())
+}
+
+pub(crate) fn normalize_numeric_offset(value: f64) -> Option<String> {
+    (value.is_finite() && (0.0..=MAX_SAFE_INTEGER).contains(&value) && value.fract() == 0.0)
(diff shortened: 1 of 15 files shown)


swernerx:
@greptileai Please review current head `02ceb31` and verify that the direct offset validation and duplicated-helper findings are addressed.

PR #438, opened 2026-07-24
fix: transform macros in Trans component attributes

### LINKED ISSUE #420: fix(transform): Trans macro inside an attribute of a replaced subtree renders empty at runtime
## Summary

`<Trans>Click <Button title={<Trans>Tooltip</Trans>} /> now</Trans>` transforms to `components={{ 0: <Button title={<Trans>Tooltip</Trans>} /> }}` — the replaced `<Trans>` subtree is never re-visited (`crates/palamedes/src/transform/visitor.rs:141-172`), while the nested-macro finder deliberately allows macros inside attributes (`:361-365`).

The inner `Trans` now binds to the **runtime** component (macro import removed, `import { Trans } from "@palamedes/react"` injected) with no `id`/`message`. The runtime `Trans` ignores children and resolves `resolvedId = id ?? message ?? ""` (`packages/react/src/index.tsx:47-65`) → it resolves message `""` and renders nothing.

Meanwhile extraction happily catalogs `"Tooltip"`, so translators translate a string the UI never shows. Verified empirically.

## Suggested fix

Either transform macros inside attribute expressions of replaced subtrees (recurse with nested replacements), or make this a hard error like other nested macros — instead of silently emitting broken output.

swernerx:
Closes #420

## Summary
- traverse independent JSX attribute/render-prop contexts before replacing an outer `<Trans>` subtree
- fold nested macro replacements into the preserved component markup instead of emitting dead macro components
- preserve React and Solid component wrapper behavior
- add regression coverage for nested `<Trans>` attributes and nested choice macros in render props

## Validation
- `CI=true RUSTUP_TOOLCHAIN=1.95 pnpm format:check`
- `CI=true RUSTUP_TOOLCHAIN=1.95 pnpm lint`

### DIFF
--- a/crates/palamedes/src/transform/messages.rs
+++ b/crates/palamedes/src/transform/messages.rs
@@ -17,6 +17,8 @@ use crate::jsx_message::{
 };
 use crate::placeholder_name::{expression_name, jsx_expression_name};
 
+use super::Replacement;
+
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub(super) struct ValueBinding {
     pub expression: String,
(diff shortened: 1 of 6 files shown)


swernerx:
@greptileai Please re-review the current head `765017403949a9d8b6d628e9c9659c0ad751be9e`; the two P2 findings are addressed with expression-container traversal/regression coverage and an explicit non-overlap invariant.

PR #439, opened 2026-07-24
fix: share runtime state across module graphs

### LINKED ISSUE #423: fix(runtime): getter and listeners are module-local while instance state is globalThis-keyed — breaks under dual module graphs
## Summary

The client instance (`Symbol.for("palamedes.runtime.clientI18n")`) and server ALS state (`Symbol.for("palamedes.runtime.serverI18nScopeState")`) are deliberately stored on `globalThis` to survive multiple module copies — but `serverI18nGetter` and `clientI18nListeners` are plain module-locals (`packages/runtime/src/index.ts:8-15` vs `:27`; `packages/runtime/src/server.ts:5,17-29`).

The package ships both ESM and CJS. If translated code resolves one copy and the app's setup code another — Next.js servers have distinct RSC and SSR module graphs; CJS `require` from instrumentation is another route — then:

- `getI18n()` on the server throws "No active server i18n instance" despite `createServerI18nScope()` having run, and
- `subscribeClientI18n` bindings (the Solid bridge, a future React bridge) never hear `setClientI18n` calls from the other copy.

The half-done globalization suggests the hazard is known but incompletely handled.

## Suggested fix

Store the getter and the listener set under the same `Symbol.for` global-state pattern — or have `getI18n`'s server branch fall back to reading `globalState[SERVER_SCOPE_STATE_KEY].active.getStore()` directly.

Also in this area: `server-unavailable.ts` re-declares `ServerI18nScope<T = unknown>` with looser generics than `server.ts:11-15` (dead duplicate type that can drift), and `"browser": "./dist/server-unavailable.mjs"` in `package.json:26` serves an `.mjs` file to `require` in browser-conditioned CJS builds.

swernerx:
Closes #423

## Summary
- store the client listener set and server getter under `Symbol.for` global runtime keys alongside the existing shared instance/scope state
- let server lookups fall back directly to the shared async-local scope for mixed module graphs
- add isolated-module regression tests for server scope resolution and client subscriptions/snapshots
- share the `ServerI18nScope` type between Node and unavailable-server entries
- provide mode-correct ESM and CJS browser/default fallbacks for `@palamedes/runtime/server`

## Validation
- `CI=true RUSTUP_TOOLCHAIN=1.95 pnpm format:check`
- `CI=true RUSTUP_TOOLCHAIN=1.95 pnpm lint`
- built ESM/CJS smoke test covering cross-format client notifications and server scope resolution

### DIFF
--- a/packages/runtime/package.json
+++ b/packages/runtime/package.json
@@ -23,8 +23,14 @@
         "import": "./dist/server.mjs",
         "require": "./dist/server.cjs"
       },
-      "browser": "./dist/server-unavailable.mjs",
-      "default": "./dist/server-unavailable.mjs"
+      "browser": {
+        "import": "./dist/server-unavailable.mjs",
+        "require": "./dist/server-unavailable.cjs"
+      },
+      "default": {
+        "import": "./dist/server-unavailable.mjs",
+        "require": "./dist/server-unavailable.cjs"
+      }
     }
   },
   "main": "./dist/index.cjs",
--- a/packages/runtime/src/index.test.ts
+++ b/packages/runtime/src/index.test.ts
@@ -83,6 +83,16 @@ describe("@palamedes/runtime", () => {
     }
   })
 
+  it("clears client listeners when the runtime is reset", () => {
+    const listener = vi.fn()
+    subscribeClientI18n(listener)
+
+    resetI18nRuntime()
+    setClientI18n(createTestI18n())
+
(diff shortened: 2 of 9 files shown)


swernerx:
@greptileai Please re-review the current head `31fba43ea04666d62511e14d7ab87c160f25cfda`. The listener reset finding is addressed, including Solid re-subscription across runtime resets and regression coverage.

PR #447, opened 2026-07-25
Fix formatting-correctness cluster: plural coercion, Date hydration, runtime choice components, canonical URL scheme (#414, #416, #417, #418)

### LINKED ISSUE #414: fix(core): missing or non-numeric plural values coerce to 0 and match =0/zero branches
## Summary

`normalizeNumericValue` (`packages/core/src/messageFormat.ts:485-492`, used at `:343` and `:468`) coerces absent/non-finite plural values to 0.

Verified by execution: `formatMessagePattern("{n, plural, =0 {none} one {one} other {# items}}", {})` → `"none"`.

## Failure scenario

A forgotten `values` entry — or `n: undefined` from an optional API field — renders the semantically loaded "you have no items" message instead of surfacing an error. Lingui/intl-messageformat throw in this case, which `createI18n` would then catch, report via `onError`, and fall back on. The silent `NaN → 0` path also affects `#` output.

The same coercion logic is duplicated in `packages/react/src/index.tsx:184-191`.

## Suggested fix

Throw (or at minimum report through `onError`) when a plural/selectordinal value is absent or non-finite; keep coercion only for numeric strings. Apply in both core and the React copy (ideally deduplicate).

### LINKED ISSUE #416: fix(react): &lt;Trans&gt; renders Date values via String(date) — timezone-dependent output and hydration mismatches
## Summary

Core's `stringifyValue` special-cases `Date` → `toISOString()` (deterministic; `packages/core/src/messageFormat.ts:527-541`). The React path (`renderVariable`, `packages/react/src/index.tsx:172-182`) does not, so `{when}` with a `Date` value renders `date.toString()`.

## Failure scenario

E.g. `"Thu Jul 24 2026 02:00:00 GMT+0000 (Coordinated Universal Time)"` on a UTC server vs `"…GMT+0200 (Central European Summer Time)"` in the user's browser → guaranteed hydration mismatch. Additionally, `i18n._` and `<Trans>` disagree for identical input.

## Suggested fix

Route bare variables in `renderVariable` through the same `stringifyValue` (export it from core), or better, format Dates with the cached locale-aware formatter.

### LINKED ISSUE #417: fix(react): runtime Plural/Select/SelectOrdinal never translate and silently differ from their macro namesakes
## Summary

The runtime choice components (`packages/react/src/index.tsx:67-94`) format the English source props directly (`formatMessagePattern(message, { value }, i18n.locale)`) — they never consult the catalog. Hand-written (non-macro) usage therefore ships untranslated in every locale, while looking identical to the macro import one path over (`@palamedes/react` vs `@palamedes/react/macro`).

Additional divergences from the macro semantics:

1. The macro transform maps `_0`/`_male` props to ICU `=0`/`=male` (`crates/palamedes/src/transform/messages.rs:61-65,219-223`), but `buildChoiceMessage` passes keys through verbatim — a runtime `<Select value="male" _male="He" other="They"/>` emits option key `_male` that can never match → always renders "They".
2. `PluralProps` lacks `offset` and exact-match props entirely.
3. Option strings containing an unbalanced `}` corrupt the synthesized pattern (e.g. `{value, select, other {a } b}}`).

## Suggested fix

Make the runtime choice components resolve through the i18n instance like `Trans` does (or document loudly that they are source-language-only), apply the same `_N → =N` normalization as the transform, and align the prop types.

### LINKED ISSUE #418: fix(core): canonicalUrl hardcodes http:// — locale suggestions and switch links downgrade HTTPS users
## Summary

Whenever a host is resolved (map/subdomain/tld modes), `canonicalUrl` returns `` `http://${hostWithPort}${canonicalPath}${search}` `` (`packages/core/src/locale.ts:519`). `suggest()` embeds this in `recommendedUrl` (`:538`, `:558`).

A production suggestion banner or locale switcher using `recommendedUrl` directly links users from an HTTPS page to `http://example.de/...` — a mixed-scheme redirect hop at best, a broken link behind HTTPS-only infra at worst.

The shipped examples all work around it by stripping the scheme (`.replace(/^https?:/, "")` in `LocaleSwitcher.tsx:27` and `SuggestionBanner.tsx:26`) — a trick every consumer has to know.

## Suggested fix

Accept a `protocol` in `LocaleControlsConfig`/`canonicalUrl` options (or derive it from a passed request URL), defaulting to `https` outside localhost — or emit protocol-relative `//host/path` URLs. Then remove the scheme-stripping workaround from the examples.

swernerx:
Block 1 of the remaining review backlog: the four correctness issues around message formatting and locale URLs. The root cause running through three of them was duplicated rendering logic between `packages/core/src/messageFormat.ts` and the React/Solid renderers — this PR consolidates it in core and has both adapters consume the shared helpers.

## Fixes #414 — missing plural values throw instead of matching `=0`/`zero`

`normalizeNumericValue` coerced absent/non-finite plural values to `0`, so a forgotten `values` entry rendered the semantically loaded "you have no items" branch. Plural/selectordinal arguments now require a present, numeric value (numeric strings still work); a violation throws a descriptive error (`Missing or non-numeric value for {n, plural, …}: received undefined.`). Inside `_()`/`getMessage()` that error is reported through `onError` and rendering falls back to the source message — same resilience contract as before. Branch selection moved into an exported `resolveChoice`, which both adapters now use, deleting the duplicated (and drifted) copies.

## Fixes #416 — `<Trans>` renders `Date` values deterministically

The React/Solid `renderVariable` used `String(date)`, producing timezone-dependent output and guaranteed SSR hydration mismatches. Bare variables now render through core's exported `stringifyValue`, so `Date` values become ISO strings — identical to `i18n._` and identical on server and client.

## Fixes #417 — runtime choice components resolve through the i18n instance

The runtime `Plural`/`Select`/`SelectOrdinal` formatted their source-language props directly. Now they:

- resolve via `getMessageNodes` exactly like `Trans`, so catalog entries keyed by the synthesized source pattern are honored;
- pass a new `reportMissing: false` metadata flag, preserving the deliberate behavior from `b04ad8d` ("Preserve direct choice component behavior") that expected catalog misses don't spam `onMissing` — the existing test locking that in still passes;
- normalize `_N`/`=N` exact-match props to ICU `=N`, mirroring `normalize_choice_option_key` in the transform, and reject invalid plural option props (previously `_male` was emitted verbatim and could never match);
- validate that the synthesized pattern round-trips its option keys, so option text with unbalanced braces raises a descriptive `RangeError` instead of silently rendering something unrelated;
- accept typed `_0`/`_1`/… props (`Record<`_${number}`, string>`) — `offset` was already supported.

## Fixes #418 — canonical URLs no longer hardcode `http://`

Host-carrying URLs from `canonicalUrl()`/`suggest()` are now protocol-relative (`//host/path`) by default — exactly what all twenty subdomain/tld example components produced by manually stripping the scheme; those workarounds are removed. A new `protocol` option on `defineLocaleControls` produces absolute URLs when needed.

## Compatibility

All four are behavior changes and documented under `## Unreleased` in CHANGELOG.md and in `docs/api/core.md` (new exports `resolveChoice`, `ResolvedChoice`, `stringifyValue`, the `MessageMetadata.reportMissing` flag, and the `protocol` option). The react README's runtime-components section is updated accordingly.

## Verification

- `pnpm build` + full `pnpm test`: 262 tests green, including new regression tests for every fix (missing-value throw + numeric-string acceptance, catalog-translated choice components, `_N` normalization, invalid-option and unbalanced-brace rejection, deterministic `Date` rendering, protocol-relative and explicit-protocol URLs).
- `pnpm check-types`, `pnpm lint` (oxlint + eslint), `pnpm format:check` — green.
- Spot-built two affected example apps (`example-tanstack-subdomain`, `example-solidstart-tld`) — green.

https://claude.ai/code/session_01TK4yk5JVr5FURvoKZhCgon

---
_Generated by [Claude Code](https://claude.ai/code/session_01TK4yk5JVr5FURvoKZhCgon)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,384 +9,358 @@
 - Custom `I18nInstance` implementations must expose an initialized
   `locale: string`; implementations that omitted the property or declared it as
   optional need to update.
+- Plural and selectordinal arguments now require a present, numeric value
+  (numeric strings still work). A missing or non-finite value throws instead of
+  silently coercing to `0` and matching `=0`/`zero` branches; inside
+  `_()`/`getMessage()` the error is reported through `onError` and rendering
+  falls back to the source message.
+- Host-carrying URLs from `canonicalUrl()` and `suggest()` are now
+  protocol-relative (`//host/path`) instead of hardcoding `http://`. Set the
+  new `protocol` option in `defineLocaleControls` for absolute URLs.
+- The runtime `Plural`/`Select`/`SelectOrdinal` components in
+  `@palamedes/react` and `@palamedes/solid` now resolve through the active
+  i18n instance (catalog entries keyed by the synthesized source pattern are
+  honored), normalize `_N` exact-match props to ICU `=N` like the macro
+  transform, and reject invalid option props or option text with unbalanced
+  braces instead of silently misrendering.
+- `<Trans>` variable values that are `Date` instances now render as
+  deterministic ISO strings (matching `i18n._`), fixing SSR hydration
+  mismatches across time zones.
 
 ## [1.6.0](https://github.com/sebastian-software/palamedes/compare/palamedes-v1.5.1...palamedes-v1.6.0) (2026-07-25)
 
-
 ### Features
 
-* **cli:** add binary plugin protocol host support ([fe519b9](https://github.com/sebastian-software/palamedes/commit/fe519b9c641da1c5989d1af0f225ebda99fe41fd))
-* **cli:** binary plugin protocol host support (ADR 002) ([c4aeb3b](https://github.com/sebastian-software/palamedes/commit/c4aeb3ba5b4b0475368ee83e905856400ff039d5))
-* **core:** initialize every i18n locale ([#432](https://github.com/sebastian-software/palamedes/issues/432)) ([57b6ecd](https://github.com/sebastian-software/palamedes/commit/57b6ecd2eb347d49d1c27b52b757c1da17ad910f))
-
+- **cli:** add binary plugin protocol host support ([fe519b9](https://github.com/sebastian-software/palamedes/commit/fe519b9c641da1c5989d1af0f225ebda99fe41fd))
+- **cli:** binary plugin protocol host support (ADR 002) ([c4aeb3b](https://github.com/sebastian-software/palamedes/commit/c4aeb3ba5b4b0475368ee83e905856400ff039d5))
+- **core:** initialize every i18n locale ([#432](https://github.com/sebastian-software/palamedes/issues/432)) ([57b6ecd](https://github.com/sebastian-software/palamedes/commit/57b6ecd2eb347d49d1c27b52b757c1da17ad910f))
 
 ### Bug Fixes
 
-* add reactive React runtime bridge ([5770cda](https://github.com/sebastian-software/palamedes/commit/5770cda2a66a3b14fbd2e643e6e53a1ee9a2450e))
-* clear shared runtime listeners on reset ([8c72577](https://github.com/sebastian-software/palamedes/commit/8c725777be95c300a077c692461ba935cfa4afa7))
-* **cli:** fail safely on extraction errors ([3f58b2f](https://github.com/sebastian-software/palamedes/commit/3f58b2fe24b7b6b176a7578c8e7c2faf904d66b4))
-* **cli:** fail safely on extraction errors ([3b4557e](https://github.com/sebastian-software/palamedes/commit/3b4557ea8445a8f8490e033a60b826768166ec95))
-* **cli:** keep watch mode alive after parse errors ([44829e4](https://github.com/sebastian-software/palamedes/commit/44829e437f76a8a863a294af2130f7e29a73ea2b))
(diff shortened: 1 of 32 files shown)


swernerx:
Response to the two Greptile P1 findings:

- **Invalid Dates abort rendering — fixed** in the latest commit: `stringifyValue` now degrades an invalid `Date` to the platform's `"Invalid Date"` string instead of letting `toISOString()` throw (regression test added). This matches the pre-consolidation React behavior.
- **"Valid" signed/fractional exact choices rejected — intentionally not changed.** The runtime validator deliberately mirrors the macro transform's `normalize_choice_option_key` rule (`crates/palamedes/src/choice.rs`), which accepts only plural categories and non-negative-integer exact matches; `_-1`/`_1.5` are a transform error in macro components too. Widening the runtime components beyond what the transform accepts would reintroduce exactly the macro/runtime divergence #417 fixed. The `Record<`_${number}`, string>` type is looser only because TypeScript cannot express "non-negative integer" in a template-literal key; the runtime error is loud and descriptive at development time.

---

PR #449, opened 2026-07-26
Config/CLI DX and quick-win sweep: loader parity, watch resilience, CLI polish, core/plugin/example hygiene (#421, #422, #425, #426, #427)

### LINKED ISSUE #421: fix(config): CLI/JS loader disagreements and validation gaps — camelCase keys silently dropped, palamedes.config.ts invisible to pmds
## Summary

Two related config-surface problems:

### 1. CLI and JS loader support different config files

`packages/config/src/index.ts:9-18` supports `palamedes.config.ts/js/mjs/cjs` in addition to data formats; `crates/palamedes-cli/src/config.rs:9-15` only supports yaml/yml/json/toml. A Next.js project configured via `palamedes.config.ts` builds fine through the next-plugin, but `pmds extract` reports "Could not find a Palamedes config. Expected one of palamedes.yaml, …" — technically true but confusing. The upward directory search also means the two sides can resolve *different* config files in a monorepo.

**Fix:** detect `palamedes.config.*` in the Rust CLI and emit a specific error ("found palamedes.config.ts, which the native CLI cannot load; use a YAML/JSON/TOML config"), or route CLI config loading through the JS host.

### 2. Silently ignored keys, no cross-field validation

`RawConfig` has no `deny_unknown_fields` (`crates/palamedes-cli/src/config.rs:63-76`); the JS side likewise ignores unknown keys (`packages/config/src/index.ts:42-54`, `140-172`, `397-450`). Lingui configs are camelCase; Palamedes data configs accept only kebab/snake_case. `sourceLocale` at least fails as a missing required field, but `pseudoLocale`, `fallbackLocales`, and `sourceReferenceRoot` written in camelCase are **silently dropped** on both sides — pseudo-locale exclusion, fallback chains, and origin-path style silently change. Neither loader validates that `pseudoLocale` is in `locales` nor that fallback-chain entries reference configured locales.

**Fix:** reject or warn on unknown top-level/catalog keys (especially camelCase spellings of known keys) and validate `pseudoLocale`/`fallbackLocales` against `locales`.

(Verified: config-rs 0.15.25 preserves map-key case for per-locale fallback maps like `pt-BR` — that specific concern is clear.)

### LINKED ISSUE #422: fix(cli): watch mode dies on the first fatal authoring error and has no debounce
## Summary

`run_extraction(config, options)?` inside the watch event loop propagates errors out of `run()` (`crates/palamedes-cli/src/main.rs:557-583`).

Fatal extraction errors (`UnsupportedMacroSyntax`, `ExplicitMessageIdsUnsupported`, `NestedMessageMacro`, `TranslationMacroOutsideFunction`) are easy to hit transiently while typing — e.g. `t({ message })` mid-edit. The watcher process exits with code 1 and stops watching; the developer keeps editing with no feedback.

Additional watch-mode gaps:

- No debouncing — every FS event triggers a full re-extraction.
- Glob sets are rebuilt per event (`should_rerun_for_path` → `build_include_set`).
- The config file itself is not watched.
- Includes outside `root_dir` are never watched.

## Suggested fix

In watch mode, print fatal errors and keep watching; add a small debounce window; watch the config file; build glob sets once.

### LINKED ISSUE #425: chore(cli): polish — pmds --version, --clean help text, report fuzzy counting, summary double-count
## Quick wins in the native CLI

- [ ] **`pmds --version`/`-V` not supported.** The clap derive sets `name`/`about` but not `version` (`crates/palamedes-cli/src/main.rs:25-31`), so only the `version` subcommand works; `pmds --version` exits 2 with "unexpected argument". Add `#[command(version)]`.
- [ ] **`--clean` help text is wrong/undocumented magic.** `main.rs:55-60` says "Remove obsolete messages from catalogs", but the implementation (`crates/palamedes/src/catalog_update.rs:256-262`, `376-381`) only deletes entries whose `obsolete-since` is older than a hard-coded 30 days and keeps undated ones. Document the grace period in the help string; consider making it configurable.
- [ ] **`pmds report` counts fuzzy entries as translated** (locked in by the test at `main.rs:1144-1167`). Overstates completeness relative to gettext conventions. Once fuzzy flags survive the update round-trip (see the translator-metadata issue), report them as untranslated or as a separate column.
- [ ] **`total_files` in the extract summary double-counts** files matched by multiple catalogs (`main.rs:324-334` sums per-catalog `file_count`).
- [ ] **Dead parameter:** `extract_from_tagged_template(..., _runtime: bool)` in `crates/palamedes/src/extract.rs:685-705` is never used — remove it.
- [ ] **Inconsistent JS config error strings/access modes:** `packages/config/src/index.ts:551-581` — async path checks `F_OK` and says "Could not find Palamedes config at X.", sync path checks `R_OK` and says "Palamedes config not found at X" (different wording, different semantics for unreadable-but-existing files). Unify.

### LINKED ISSUE #426: chore(core/react): hygiene — formatter caching gaps, parser error quality, typing and dead-code cleanups
### Performance

- [ ] `packages/core/src/messageFormat.ts:474-480` — `selectChoice` constructs an `Intl.PluralRules(undefined, …)` unconditionally and then usually discards it for a second locale-bound instance; `Intl.PluralRules` is never cached (unlike number/date formatters). Cache it; delete the dead first allocation.
- [ ] `packages/core/src/messageFormat.ts:523-525` — `replacePound` builds a fresh `Intl.NumberFormat` per text node, bypassing `numberFormatCache`. Plural-heavy lists pay formatter construction per row.
- [ ] `packages/core/src/messageFormat.ts:38` — `parseCache` is unbounded while formatter caches are capped at 64; since `i18n._(rawPattern)` accepts arbitrary strings, an app interpolating dynamic fallback patterns grows it forever on a long-lived server. Apply the same cap.

### Error quality

- [ ] `packages/core/src/messageFormat.ts:288-293` — parser errors (`Expected "," while parsing message pattern.`) carry no pattern excerpt, index, or actually-found character; for the stray-`}` case the message is misleading. Include position + snippet (they flow into `onError` telemetry).

### Typing / dead code

- [ ] `packages/core/src/index.ts:32` vs `:122-125` — `PalamedesI18n.locale` is typed as a writable optional property but implemented getter-only; `i18n.locale = "de"` type-checks but throws at runtime. Type it `readonly locale: string | undefined`.
- [ ] `packages/react/src/index.tsx:146` — unreachable `return []` after an exhaustive switch; `:37-39` — pointless `{} & ChoiceComponentProps` intersections.
- [ ] `packages/react/src/index.tsx:178` — `cloneElement(value, {})` omits a `key` for element-typed values → React key warnings for `values={{ icon: <Icon/> }}`.
- [ ] `packages/react/src/index.tsx:197` — undocumented `export { Fragment }` re-export of React's Fragment (presumably for transform output); document or remove.
- [ ] `packages/palamedes/src/index.ts` — `export {}` in a package that isn't published with `dist` (`files: ["README.md", "bin/"]`); dead source file for the placeholder package.
- [ ] `packages/runtime/src/server.ts:39-43` — add a code-comment warning on `activate()` that it must be called inside a per-request async context (README covers it; copy-pasters don't read READMEs).

### LINKED ISSUE #427: chore(plugins/examples): hygiene — loader warnings channel, stale changelogs, example correctness fixes
### Plugins

- [ ] **Loader warnings bypass the bundlers' diagnostics channel.** `packages/next-plugin/palamedes-po-loader.cjs:45` and `packages/vite-plugin/src/index.ts:243` use `console.warn` instead of `this.emitWarning(...)` / `this.warn(...)`. Warnings (missing translations, ICU diagnostics) don't show in the Next overlay or vite's dedup'd output and repeat on every rebuild.
- [ ] **Stale per-package changelogs:** `packages/next-plugin/CHANGELOG.md` and `packages/vite-plugin/CHANGELOG.md` top out at 0.6.0 (2026-05-22) while the packages ship 1.5.x — release automation isn't writing them; npm users see two months of missing history including the 1.5.0 deferred-messages removal.
- [ ] **Duplicate catalog-module logic in JS and Rust:** `packages/transform/src/catalogLoader.ts` reimplements `create_catalog_module_result` (`crates/palamedes-node/src/catalog.rs`) including the pseudo guard; the plugins use only the Rust path. Documented public subpath, so not dead code — but a drift hazard. Consider deprecating or generating both from one source.
- [ ] **`parseAcceptLanguage` treats `q=0` as acceptable** (`packages/core/src/locale.ts:114-136`) — RFC-wise `de;q=0` means "not acceptable" but can still be selected if it's the only supported match.

### Next.js examples (the migration blueprint — worth getting right)

- [ ] **Route example: unknown `/:locale` renders fallback content with HTTP 200** (`examples/nextjs-route/src/app/[locale]/page.tsx:42-46`): `getRouteLocale("xx")` falls back to Accept-Language and renders normally at `/xx` — every garbage segment becomes an indexable duplicate page. `suggest()` then builds `replaceLocaleInPath("/xx", "de")` → `/de/xx` (segment prepended since `xx` isn't a locale), a 404 that also renders. Call `notFound()` when `params.locale` fails `locales.isLocale`; consider `generateStaticParams`.
- [ ] **Hardcoded `<html lang="en">`** in `examples/nextjs-route`, `nextjs-subdomain`, `nextjs-tld` (`src/app/layout.tsx:11` each) — only the cookie example sets `lang` from the resolved locale. Wrong for screen readers and SEO.
- [ ] **Dead conditional** in `examples/nextjs-route/src/lib/i18n.server.ts:17-18`: both branches produce the same string whenever `paramsLocale` is defined.
- [ ] **ProofPanel caption mismatch:** renders `{when, date, medium}` but the code caption claims `{when, date, full}` (`examples/nextjs-cookie/src/components/ProofPanel.tsx:88,91`; present in all four examples).

swernerx:
The remaining DX pair plus all three quick-win collections from the review backlog, in one PR as requested. Four commits, one per issue cluster.

## Fixes #421 — config loader parity and validation

- `pmds` now reports `palamedes.config.ts`/`.js` files with a specific, actionable error ("the native CLI cannot load JavaScript/TypeScript configs; create a palamedes.yaml…") instead of the generic not-found.
- Both loaders (Rust and JS) reject camelCase spellings of known data-config keys (`sourceLocale`, `pseudoLocale`, `fallbackLocales`, `sourceReferenceRoot`) with a kebab-case hint instead of silently dropping them; other unknown top-level keys warn.
- Fallback-locale chains must reference configured locales (hard error). A `pseudo-locale` outside `locales` warns — a hard error would contradict the documented "values outside locales are ignored" behavior (and the existing test fixture that relies on it).
- The async/sync config-not-found paths now share `F_OK` semantics and one error message (the #425 unification item).

## Fixes #422 — watch mode survives editing

Every extraction failure — including fatal authoring errors like a half-typed `t({ message })` — prints a warning and watching continues. Plus: event debouncing (150 ms quiet window), glob sets built once per config generation instead of per event, include roots outside `root_dir` are watched, and `palamedes.yaml` itself is watched and reloaded on change (bad reloads keep the previous config and continue).

## Fixes #425 — CLI polish

- `pmds --version`/`-V` work (`#[command(version)]`); `pmds version` unchanged.
- `--clean` help documents the 30-day obsolete grace period and points at `--force-clean`.
- `pmds report` counts fuzzy PO entries as untranslated and shows them in a new `fuzzy` column (gettext semantics; the fuzzy flags are read from the PO storage directly since the parse API doesn't expose them). The test that locked in fuzzy-as-translated now asserts the corrected semantics. Documented as a compatibility note — completeness percentages can drop.
- The extract summary counts each source file once across overlapping catalogs; the dead `_runtime` parameter is removed.

## Fixes #426 — core/react hygiene

- `Intl.PluralRules` cached like the other formatters (the old path allocated one per choice resolution plus a discarded extra); `replacePound` reuses the cached number formatter; the parse cache is bounded at the same 64-entry limit as the formatter caches.
- Parser errors carry the failure index, the found character, and a ±20-char pattern excerpt.
- Solid drops the unreachable `return []`; the react README documents the `Fragment` re-export; `ServerI18nScope.activate()` carries the async-context warning in code; the placeholder package's `src/index.ts` documents why it exists (it feeds the workspace `tsc --noEmit` sweep). The `locale` typing, `{} &` intersections, and `cloneElement` key items were already fixed in earlier PRs.

## Fixes #427 — plugin/example hygiene

- Vite/Next `.po` loaders report diagnostics through `this.warn`/`emitWarning` — deduplicated, visible in the overlays, no longer repeated on every rebuild.
- `parseAcceptLanguage` treats `q=0` as "not acceptable" per RFC 9110.
- **Next route example restructured:** `/` negotiates via middleware, `app/[locale]/layout.tsx` is the root layout with `generateStaticParams`, `notFound()` for unknown segments (no more indexable `/xx` fallback pages), and a correct `html lang`. The subdomain/tld examples resolve `lang` from the request host; the dead pathname conditional is gone.
- All twenty ProofPanel captions now name the format they actually render (`{when, date, medium}`); the `{when, date, full}` claim in examples/README stays valid via TicketPanel, which really renders `full`.
- The 18 per-package `CHANGELOG.md` files (frozen at the 0.6.x per-package release era, not shipped to npm, not written by release-please's single-root setup) now point at the root changelog.
- The JS/Rust catalog-module duplication item: the JS twin was aligned semantically in #444 (`resolvedLocaleChain`); it stays as a documented public subpath — generating both from one source would be a separate infrastructure change and isn't attempted here.

## Verification

- `cargo test --workspace` (4 suites, incl. new config tests for the JS-config error, camelCase rejection, and fallback validation, and the corrected fuzzy report test), `cargo clippy` clean, `cargo fmt --check` clean.
- `pnpm build` + full `pnpm test` (265 tests), `pnpm check-types`, `pnpm check:native-types`, `pnpm lint`, `pnpm format:check` — green.
- All four Next.js examples build; the restructured route example passes `verify-examples` smoke (SSR, Accept-Language, route rendering, switch targets).
- `docs/cli.md` updated (watch behavior, fuzzy column, `--version`, configuration-error section); CHANGELOG compatibility notes added for every behavioral change.

https://claude.ai/code/session_01TK4yk5JVr5FURvoKZhCgon

---
_Generated by [Claude Code](https://claude.ai/code/session_01TK4yk5JVr5FURvoKZhCgon)_


swernerx:
Both Greptile findings on the watch-reload path are addressed in the latest commit: config reloads now diff the watch-root set — removed roots are unwatched (P2), and a failed registration for a newly configured root prints a warning naming the path instead of silently reporting a successful reload (P1). Clippy, fmt, and the CLI test suite are green.

---

PR #464, opened 2026-07-27
fix: address the post-1.5 re-review findings (apostrophe parity, adapter resilience, cache hardening, docs)

swernerx:
## Summary

Implements all confirmed findings from the follow-up production-readiness review of the post-1.5 fix wave, in four logical commits:

1. **Rust — ICU apostrophe parity + extraction-cache hardening** (`fix(core,cli)`)
   - New shared `icu_text` module: the extractor and transform now escape authored apostrophes (`'` → `''`) at every literal ingestion site, so the runtime's lenient ICU quoting renders them literally — previously `` t`L'${title}` `` swallowed the placeholder. Extractor msgids and transform runtime messages stay byte-identical (cross-lock test).
   - Compile/audit/pseudo-localization switch from `IcuSyntaxPolicy::RuntimeLiteralApostrophes` to `Strict` with catalog-text canonicalization at load, so `pmds audit` now flags quote-swallowed placeholders instead of staying green while the runtime breaks.
   - Watch mode rebuilds the extraction cache on stamp-relevant config reloads (was serving stale origins/scopes); the library API also resets a cache whose stamp no longer matches the request.
   - Cache racy window measured from read start instead of insert time (coarse-mtime filesystems).
   - `extract-threads`/`extract-cache` config keys no longer trigger the false "Ignoring unknown config key" warning; camelCase spellings get the kebab-case hint.
   - Failed-file warnings/count deduped across overlapping catalogs; watch debounce drain bounded at 2s; unbuildable watch globs warn instead of silently disabling a catalog.
   - Coverage debts closed: `retain_paths` retention tests, cache-warm vs `--no-cache` byte-parity test, watch-reload regression test.
2. **JS — adapter render resilience + hardening** (`fix(core,react,solid)`)
   - `<Trans>`/`<Plural>`/`<Select>`/`<SelectOrdinal>` now follow core's catch → `onError` → source-fallback contract via a new `PalamedesI18n.reportError()` channel instead of throwing out of the component render on unresolvable catalog messages (translator-triggerable crash).
   - Choice options are null-prototype + `Object.hasOwn` lookups — select values like `"toString"`/`"valueOf"` no longer reach `Object.prototype` and crash the render.
   - React reconnects reactivity after `resetI18nRuntime` (persistent client-store bridge, mirroring the earlier Solid fix).
   - Parse cache gets its own 1024-entry LRU bound (was sharing the 64-entry FIFO with the Intl caches and thrashing); adapters reuse core's cached `Intl.NumberFormat` for `#` replacement; choice components without any usable string option throw a descriptive error instead of rendering `""`.
3. **next-plugin** (`fix(next-plugin)`): webpack `.po` rule excludes `node_modules` (mirrors Turbopack's `not: "foreign"`); user Turbopack loader-shorthand arrays are normalized into rule configs before the Palamedes rule is appended; webpack vs Turbopack `include`/`exclude` matching semantics documented.
4. **Docs** (`docs:`): llms bundles updated to the shipped `--threads`/extraction-cache surface (they still steered assistants to a deleted ADR and away from a real feature); configuration guide no longer claims the JS loader reads the extract keys; ICU apostrophe quoting and `MessageLiteralNode` renderer handling documented; plural `offset` semantics documented across core/React/Solid references; new `reportError`/`replacePoundPlaceholders` exports listed; CLI README gains `--threads`/`--no-cache`; `referenceScopes` default stated; `canonicalUrl` `protocol` option cross-referenced from the locale-strategy guides.

**Migration note (apostrophe escaping):** compiled IDs change only for messages whose authored text contains an apostrophe (`don't` → msgid `don''t`). On the next `pmds extract`, old entries become obsolete and the escaped forms are added; existing translations for those messages need to be carried over. Compile/audit canonicalize legacy catalog texts at load, so already-compiled catalogs line up with bundles built by the new transform — but an old-transform bundle against a new core needs a rebuild.

## Issue

Follow-up to the re-review of the fixes for #402–#427; no dedicated issues exist for these findings — they were reported in the review session that produced this PR.

## Validation

- `cargo fmt --all --check`, `cargo test --workspace` (all green, includes the new regression tests), clippy clean per the implementation runs
- `pnpm --filter @palamedes/{core,runtime,react,solid,next-plugin,vite-plugin,config} test` — 173 tests green
- `pnpm format:check`, `pnpm lint:oxlint`, `pnpm lint:eslint` — clean
- Documented quoting/offset semantics were verified against the actual runtime behavior, not the spec

## Follow-up

- Ferrocat has no lenient-quoting policy; parity relies on canonicalizing catalog texts at load (documented in `runtime_icu_options`). A native lenient policy upstream would simplify this.
- The remix loader still caches its config per process and warns via `console.warn` (module hooks expose no watch/warn channel) — known limitation, out of scope here.
- No NodeNext-ESM lane in `scripts/check-published-types.mjs` yet (bundler-ESM, NodeNext-CJS, and node10 are covered).

https://claude.ai/code/session_01Y6YSHuSe7G5JTdhoARCXMs

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y6YSHuSe7G5JTdhoARCXMs)_

### DIFF
--- a/crates/palamedes/src/icu_text.rs
+++ b/crates/palamedes/src/icu_text.rs
@@ -124,6 +124,32 @@ pub(crate) fn canonicalize_catalog_quoting(catalog: &mut ferrocat::ParsedCatalog
     }
 }
 
+/// Derives the compiled lookup id for a message text.
+///
+/// Catalog compilation canonicalizes every catalog text on load (see
+/// [`canonicalize_catalog_quoting`]) and only then hands the `msgid` to
+/// Ferrocat's key derivation, so a compiled artifact is always keyed by the
+/// hash of the *canonical* text. Every other derivation of the same id — the
+/// lookup key the transform embeds in `getI18n()._("…")`, the ids it reports as
+/// `compiled_ids`, extractor-side parity checks — has to go through this helper
+/// so it hashes the same string.
+///
+/// Without it the raw-ICU authoring surfaces (descriptor string literals and
+/// the `<Trans message>` attribute, where authors write ICU quoting themselves
+/// and neither extractor nor transform escapes anything) hash a text the
+/// compiled catalog never contains, and the runtime silently falls back to the
+/// source message.
+///
+/// Only the message text is canonicalized: `context` is a plain disambiguation
+/// string, not ICU, and catalog loading leaves it untouched as well.
+///
+/// Because [`canonicalize_runtime_quoting`] is a fixed point, already-escaped
+/// authored text (`don''t`) keeps exactly the id it had before, so ids for the
+/// escaping paths are unchanged.
+pub(crate) fn compiled_message_key(message: &str, context: Option<&str>) -> String {
+    ferrocat::compiled_key(&canonicalize_runtime_quoting(message), context)
+}
+
 fn canonicalize_in_place(value: &mut String) {
     if value.contains('\'') {
         *value = canonicalize_runtime_quoting(value);
@@ -220,7 +246,7 @@ fn split_argument_part(input: &str) -> Option<(&str, &str)> {
 
 #[cfg(test)]
 mod tests {
-    use super::{canonicalize_runtime_quoting, escape_icu_literal};
+    use super::{canonicalize_runtime_quoting, compiled_message_key, escape_icu_literal};
 
     #[test]
     fn doubles_authored_apostrophes() {
@@ -267,6 +293,32 @@ mod tests {
         assert_eq!(
             canonicalize_runtime_quoting("{count, select, other {'#' items}}"),
             "{count, select, other {''#'' items}}"
+        );
+    }
+
+    #[test]
+    fn compiled_message_key_hashes_the_canonical_text() {
+        // The raw and the canonical spelling of the same message collapse onto
+        // one id, which is what makes a raw-ICU descriptor reachable in a
+        // compiled catalog.
+        assert_eq!(
+            compiled_message_key("Don't greet {name}", None),
+            ferrocat::compiled_key("Don''t greet {name}", None)
+        );
+
+        // Escaped authored text is a fixed point, so its id never moves.
+        for message in ["don''t", "L''{title} est prêt", "Hello {name}"] {
+            assert_eq!(
+                compiled_message_key(message, None),
+                ferrocat::compiled_key(message, None),
+                "message: {message}"
+            );
+        }
+
+        // Context is a plain disambiguation string and stays untouched.
+        assert_eq!(
+            compiled_message_key("Hello", Some("don't touch")),
+            ferrocat::compiled_key("Hello", Some("don't touch"))
         );
     }
 
--- a/crates/palamedes/src/icu_text.rs
+++ b/crates/palamedes/src/icu_text.rs
@@ -1,0 +1,290 @@
+//! Shared helpers that keep authored ICU text and the JS runtime parser aligned.
+//!
+//! The runtime parser (`packages/core/src/messageFormat.ts`) implements lenient
+//! ICU apostrophe quoting:
+//!
+//! * `''` renders a single literal apostrophe.
+//! * A single `'` starts a quoted literal only when the next character is `{`,
+//!   `}`, or — inside a plural/selectordinal branch — `#`.
+//! * Every other apostrophe is ordinary text (`don't` renders verbatim).
+//! * An unterminated quote auto-closes at the end of the pattern.
+//!
+//! Two things follow from that, and both live here so the extractor, the
+//! transform, and catalog validation cannot drift apart:
+//!
+//! * [`escape_icu_literal`] escapes authored literal segments (template literal
+//!   quasis, JSX text, choice option strings) so an authored apostrophe never
+//!   turns into a quote that swallows the following placeholder.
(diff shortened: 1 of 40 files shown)


crates/palamedes/src/icu_text.rs
```diff
@@ -0,0 +1,290 @@
...
+/// escaping robust across segment boundaries: an authored segment ending in `'`
+/// followed by a `{placeholder}` cannot form a quote, because the segment is
+/// escaped before the placeholder is appended.
+pub(crate) fn escape_icu_literal(text: &str) -> String {
+    if text.contains('\'') {
+        text.replace('\'', "''")
+    } else {
+        text.to_owned()
+    }
+}
```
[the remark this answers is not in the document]
swernerx: This is deliberate, not a gap: a descriptor with a string-literal `message` is the raw-ICU authoring surface. Placeholders there are written literally — `t({ message: "Hello {name}" })` is long-standing, test-covered usage (`extract.rs`, "extracts descriptor messages with comments") — so the author owns the ICU semantics, including quoting. Auto-escaping would corrupt exactly the inputs this surface exists for: `It''s {name}` would become `It''''s {name}` (rendering `It''s`), and an intentionally quoted `L'{title}` would flip into an active placeholder. `L'{title}` rendering the brace literally is correct lenient-ICU behavior for a raw pattern, identical to writing the same string in a `.po` file.

Both pipelines already agreed on this (extract `extract_from_descriptor_call` and transform `transform_descriptor_object` both pass string literals through raw, so compiled IDs stay aligned), but the distinction was unlocked and undocumented. Addressed in 7893a6f:

- new parity test `descriptor_string_literals_stay_raw_icu_on_both_sides` locking raw string-literal descriptors and auto-escaped template-literal descriptors byte-for-byte across extract and transform,
- the escaping exception documented in `docs/api/core.md` ("Quoting and Literal Text") and `packages/core/README.md`.

---

crates/palamedes/src/transform/tests.rs
```diff
@@ -334,6 +334,85 @@ const richChoice = <Plural value={props.quantity()} one="# task" other="# tasks"
...
+            .map(|message| message.message.as_str())
+            .collect::<Vec<_>>(),
+        vec!["L'{title} est prêt", "It''s {name}", "l''{item}"]
+    );
+
+    let extracted_ids = extracted
+        .iter()
+        .map(|message| compiled_key(&message.message, message.context.as_deref()))
+        .collect::<Vec<_>>();
+    assert_eq!(transformed.compiled_ids, extracted_ids);
```
swernerx, line 428: This only locks extractor ↔ transform parity, but the production boundary also includes catalog compilation. `compile_catalog_artifact` canonicalizes `Don't greet {name}` to `Don''t greet {name}` before deriving the key. On this head I reproduced transform key `UU28cSkwwyE` versus compiled catalog key `Lzr5NTlRSX0`, so the runtime cannot reach the translation. Please extend this regression test through `compile_catalog_artifact` (and cover `<Trans message="…">`) and require the transform lookup ID to exist in the compiled artifact.

swernerx (commented):
Merge blocker: the new raw-ICU descriptor exception still breaks the end-to-end lookup key contract. The extractor and transform agree with each other, but catalog loading canonicalizes every msgid before compilation. A descriptor such as `t({ message: "Don't greet {name}" }, { name })` therefore emits transform key `UU28cSkwwyE`, while `compile_catalog_artifact` emits catalog key `Lzr5NTlRSX0`; the translation is unreachable and source fallback is rendered. `<Trans message="…">` follows the same unescaped path and is not covered by the documented descriptor exception. Please align lookup-key generation with catalog canonicalization (while preserving whatever raw runtime fallback semantics are intended) and add a Transform → Catalog Compile cross-lock test for descriptor strings and message attributes.

crates/palamedes/src/transform/tests.rs
swernerx: Confirmed and fixed in 81f2e3d. The failing cross-lock test written first reproduced your keys exactly (`UU28cSkwwyE` from the raw text vs `Lzr5NTlRSX0` from the canonicalized catalog), and `<Trans message="…">` was indeed the same raw path — `jsx_attribute_string_value` only decodes entities.

What changed:

- New shared `compiled_message_key(message, context)` in `icu_text.rs` — canonicalize, then hash. All five key-derivation sites in `transform/runtime.rs` route through it; the embedded runtime message and the extracted msgids stay byte-identical, only the hashed input changed. Compile/audit already keyed by canonical text, so all three stages now agree, and legacy-catalog key mapping is preserved.
- `catalog_artifact::tests::transform_lookup_ids_match_compiled_catalog_keys`: transform + extract over a raw descriptor, `<Trans message="…">`, and an escaped tagged template → PO catalog from the extractor's msgids → `compile_catalog_artifact` → every transform lookup id must exist in the artifact and resolve to the right translation.
- Escaped authored paths keep their previous ids (idempotence verified via the existing parity tests, which pass untouched); the raw-ICU surfaces move onto the catalog's keys — a bundle built from earlier commits of this branch needs a rebuild.
- `<Trans message>` added to the documented raw-ICU exception in `docs/api/core.md` and the core README.

---

PR #476, opened 2026-07-28
refactor(cli): split commands into modules behind a Command trait

swernerx:
## Why

`crates/palamedes-cli/src/main.rs` had grown to 1862 lines holding every command's options, pipeline, rendering, and tests. This splits it into one module per namespace and gives each command an explicit contract.

This came out of a discussion about whether the new native plugin system (ADR 018) should be used to modularize large internal commands. The conclusion was no — moving built-ins behind the protocol would freeze their interface into a versioned public contract, put Node in the hot path for every `extract`, and give up the guarantee that a broken plugin can never affect them (ADR 017). Git makes the same split: core commands in one binary, subprocess dispatch reserved for third-party extensions. So the modularity happens at module level, where it costs nothing.

## What

`main.rs` is now 21 lines. The largest remaining file is `commands/extract/mod.rs` at 565.

```
src/
  main.rs                       21   parse + error reporting
  cli.rs                        55   clap surface + dispatch table
  command.rs                    83   Command trait + Context
  error.rs                      51   CliError
  config.rs                    713   unchanged
  commands/
    mod.rs                      43   helpers shared by >1 command
    extract/{mod,watch,sources,cache}.rs   565/346/136/79
    catalog/{convert,merge,mod}.rs         192/106/23
    report.rs                  342
    audit.rs                   113
    version.rs                  22
```

### The trait

```rust
fn run(&self, context: &Context) -> Result<Self::Output, CliError>;
fn render(&self, output: &Self::Output) -> Result<(), CliError>;
fn verdict(&self, output: &Self::Output) -> Result<(), CliError> { Ok(()) }
```

`verdict` runs **after** `render`. That makes an invariant which previously lived only in statement order enforceable by the type: `--fail-on` and `--fail-if-below` decide the exit code but can never suppress the report that explains the decision.

`Context` carries the working directory and config loading, replacing five scattered `std::env::current_dir().expect("current dir")` calls and making commands testable without `chdir`. Built-ins keep owning config discovery, because `catalog merge` runs as a Git merge driver in arbitrary worktrees and `version` answers before any project exists.

The shape deliberately mirrors a binary plugin command (ADR 018), so a command that ever does move behind that protocol would swap its context and renderer rather than its body.

## Behavior

Unchanged. Beyond the test suite, I built a binary from the pre-split source and diffed 19 invocations against it — every command, both JSON modes, the error paths, and the `PALAMEDES_TIMING_JSON` marker the e2e benchmark parses. Identical output and exit codes throughout.

One internal restructure: `catalog convert` now returns a report rendered in `render` instead of printing mid-loop. The emitted text is unchanged.

## Validation

- `cargo clippy --workspace --all-targets --locked -- -D warnings` — clean
- `cargo test --workspace --locked` — 225 passed
- `cargo fmt --all --check` — clean

## Found in passing, not fixed here

Two pre-existing bugs, both reproducing identically on the pre-split binary, both in subcommands that have no test coverage:

- `catalog convert --to fcl` fails in both modes with ``FCL catalog must start with the `%FCL1` header`` — it appears to read the PO input using the target format.
- `catalog merge` writes its output file correctly but still exits 1 with ``I/O error for ``:`` (note the empty path, originating in ferrocat). A merge driver exiting non-zero looks like a failed merge to `git merge`, so this one matters.

Tracked separately rather than folded in here.


swernerx:
Review result: the module split itself is clean and the moved implementations look behaviorally equivalent, but I found one concrete regression and one smaller testability gap.

1. **`pmds version` now panics when the current directory is unavailable.** [`Cli::execute()`](https://github.com/sebastian-software/palamedes/blob/a60ec9d6bf63fb83fc8a7ec988c2921334315143/crates/palamedes-cli/src/cli.rs#L41) constructs `Context::from_env()` before dispatching, and [`from_env()`](https://github.com/sebastian-software/palamedes/blob/a60ec9d6bf63fb83fc8a7ec988c2921334315143/crates/palamedes-cli/src/command.rs#L37-L41) calls `current_dir().expect(...)` even for `VersionCommand`, which does not use the context. I reproduced this from a deleted cwd: the `main` binary still prints the version and exits 0, while this PR panics with `current dir: ... NotFound`. The same unconditional lookup also affects `catalog merge --source-locale ...`, which previously did not inspect the cwd. Please create/resolve the context only for commands that need it, or make cwd resolution lazy and fallible.

2. **Non-blocking:** the advertised “testable without `chdir`” benefit is not currently exposed. `Context::cwd` is private and `Context::from_env()` is the only constructor, so command-module tests cannot inject a temporary cwd; the existing tests continue to bypass `Command::run()` and call inner helpers directly. A `Context::new`/`from_cwd` constructor would let tests cover the new command contract itself.

Validation on the PR head: `cargo fmt --all --check`, `cargo +1.95 clippy --workspace --all-targets --locked -- -D warnings`, and `cargo +1.95 test --workspace --locked` (225 tests) all pass. All GitHub CI jobs are green.

swernerx:
Both points confirmed and fixed in 1fb0cb6.

**1. The `version` regression is real, and my verification was at fault.** I reproduced it exactly as described: on the PR head `pmds version` from a deleted cwd panics at `command.rs:39`, while the pre-split binary prints the version and exits 0. Same for `catalog merge --source-locale`. My 19-invocation behavioral diff missed it because every one of those invocations ran from a valid working directory — the diff proved the commands do the same thing, not that they need the same environment to start.

Fixed by making the lookup lazy and fallible, as suggested. `Context` now holds `Option<PathBuf>`, `from_env()` stores `None`, and the process lookup happens inside `load_config` — so a command that never reads configuration never triggers it. A failed lookup is now `CliError::CurrentDir` rather than a panic, which also improves `extract`/`audit`/`report`: those panicked on the pre-split code too, and now report the problem instead.

Verified across all three binaries from a deleted cwd:

| | `version` | `catalog merge --source-locale` | `report` |
|---|---|---|---|
| pre-split `main` | ok | ok | panic |
| PR head `a60ec9d` | **panic** | **panic** | panic |
| fixed `1fb0cb6` | ok | ok | clean error |

**2. Fixed too, since it was cheap.** `Context::with_cwd` (test-only, so it does not become dead code in release builds) lets a test drive a command through its full contract. Used it to add three tests:

- `command::tests::from_env_defers_resolving_the_working_directory` — pins the root cause directly, so the eager lookup cannot come back.
- `commands::report::tests::threshold_failures_come_after_the_report_is_rendered` — drives `run` → `render` → `verdict` and asserts the report exists before the threshold failure. This is the invariant the trait split exists for, and it was previously untested. Passes `config: None`, so it also covers config discovery from the injected directory.
- `commands::report::tests::out_of_range_thresholds_fail_before_running`.

CLI crate is now at 25 tests, workspace at 229. `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, and `cargo test --workspace --locked` all pass.

One note on the toolchain: you validated with `+1.95`, I used `+1.97.1` — the default toolchain in my environment is a `1.95.0-nightly` that cannot compile `oxc_data_structures` (`cold_path` not yet stable). Worth knowing if that nightly is pinned anywhere shared.

PR #478, opened 2026-07-28
Add first-class MDX message extraction and Vite compilation

### LINKED ISSUE #477: Add first-class MDX message extraction and Vite compilation
## Summary

Add first-class `.mdx` support to Palamedes so natural-language Markdown and MDX content is extracted and translated automatically, without requiring authors to wrap every paragraph or heading in `Trans`.

The implementation should use FerroMark's semantic MDX event stream for syntax recognition, while Palamedes remains responsible for message boundaries, rich-text placeholders, message identity, catalog integration, and runtime lowering.

## Goals

- Treat MDX files as first-class extraction and build inputs.
- Use one shared native analysis pipeline for CLI extraction and Vite compilation.
- Reuse the existing source-string-first message identity and rich JSX message model.
- Keep the Vite integration thin: file loading, caching, HMR, and module hand-off belong in TypeScript; translation semantics belong in the Rust core.
- Continue using PO/FCL catalogs and the existing runtime fallback behavior.

## Translation semantics

Create one message per semantic content unit, including:

- headings
- paragraphs
- paragraphs within list items and blockquotes
- table cells
- image alternative text where applicable
- configured user-facing front-matter or JSX attribute values

Inline structure should remain part of the surrounding rich message:

- emphasis, strong text, links, and inline components become component placeholders
- JSX expressions become value placeholders
- link labels are translatable; URLs are not
- inline code is protected from translation
- fenced and indented code blocks are not extracted
- JSX tag names, attribute names, ESM, imports, exports, and JavaScript source are not translated
- tag-only flow constructs do not create messages
- authors have an explicit ignore/opt-out mechanism for exceptional content

Whitespace normalization and placeholder naming must be deterministic so extraction and compilation produce identical message identities.

### Rust core

Add an MDX analysis module backed by FerroMark's `mdx` feature:

1. Parse the source into source-ranged semantic MDX events.
2. Group block and inline events into translation units.
3. Convert Markdown formatting, links, JSX, and expressions into Palamedes rich-message parts.
4. Return extracted messages, diagnostics, source locations, and the information required to generate the runtime module.
5. Generate or retain enough range information for native source maps.

The Node boundary should expose a coarse workflow operation rather than many parser-level calls.

### CLI extraction

- Include `.mdx` in directory expansion and configured source matching.
- Route MDX files through the same native analyzer used by the build transform.
- Preserve comments and exact source locations in extracted catalog entries.
- Ensure catalog updates remain unchanged regardless of whether a message originated in TSX or MDX.

### Vite integration

Add an early `palamedes:mdx` plugin stage that:

- recognizes configured `.mdx` files
- calls the native MDX workflow
- emits a framework-compatible module using the existing Palamedes runtime contract
- passes ESM and embedded expressions through unchanged
- participates in Vite caching and HMR
- invalidates affected modules when relevant catalog or configuration files change
- composes correctly with the existing macro transform and React/Solid toolchains

The implementation must account for the currently supported Vite version range. If module-type handling requires a narrower version, that constraint must be explicit and tested.

## Acceptance criteria

- `pmds extract` discovers messages in representative MDX headings, paragraphs, lists, blockquotes, tables, links, inline JSX, and expressions.
- Importing an MDX module through Vite renders source-locale fallback content and catalog translations correctly.
- Extraction and Vite compilation derive identical message IDs and placeholder structures from the same source.
- React and Solid integration fixtures cover the supported module contract.
- Code, URLs, ESM, tag names, and attribute names are never extracted.
- Configured translatable attributes/front-matter fields and explicit ignore behavior are covered by tests.
- Invalid structural MDX reports source-ranged diagnostics.
- Changed modules include usable source maps.
- Catalog changes update the rendered module through HMR.
- Existing TS/TSX extraction, transforms, PO loading, and runtime behavior remain unchanged.
- Architecture and authoring behavior are documented with an end-to-end example.

## FerroMark foundations

- [ferromark#85 — Add source-spanned MDX segments](https://github.com/sebastian-software/ferromark/issues/85)
- [ferromark#86 — Add strict MDX parsing with structured diagnostics](https://github.com/sebastian-software/ferromark/issues/86)
- [ferromark#87 — Recognize inline JSX and expressions in MDX content](https://github.com/sebastian-software/ferromark/issues/87)
- [ferromark#88 — Promote container-local tag-only MDX paragraphs to flow events](https://github.com/sebastian-software/ferromark/issues/88)
- [ferromark#89 — Expose a narrow typed MDX event stream for compiler consumers](https://github.com/sebastian-software/ferromark/issues/89)

## Architecture constraints

This work should follow:

- [ADR-002: Rust-First Core with Thin Host Adapters](https://github.com/sebastian-software/palamedes/blob/main/adr/002-rust-first-core-with-thin-host-adapters.md)
- [ADR-008: Framework Adapter Architecture](https://github.com/sebastian-software/palamedes/blob/main/adr/008-framework-adapter-architecture.md)
- [ADR-014: Native Transform Source Maps](https://github.com/sebastian-software/palamedes/blob/main/adr/014-native-transform-source-maps.md)

## Non-goals

- Reimplementing Markdown or MDX parsing in Palamedes.
- A Remark/Rehype-compatible plugin system.
- Ardo-specific module contracts or site integration.
- Remote or model-backed translation calls during Vite compilation.
- Translating arbitrary JavaScript strings.
- Full mdxjs compiler compatibility beyond the documented FerroMark syntax contract.

swernerx:
## Summary

- add a FerroMark-backed native MDX analysis pipeline shared by catalog extraction and compilation
- derive semantic translation units for headings, paragraphs, lists, blockquotes, table cells, images, configured JSX attributes, and scalar frontmatter
- preserve rich markup and MDX expressions as deterministic component/value placeholders while excluding URLs and code from translator input
- harden generated JSX for quoted attributes, React `className`, fenced-code language metadata, repeated expressions, comments, ignore directives, and unsupported footnotes
- expose typed N-API analysis and route `.mdx` through the CLI, extraction cache, and public extractor
- make React and Solid MDX pass through real Vite framework pipelines, with React JSX module typing and explicit Solid `.mdx` handling
- add a standalone three-page React/Vite MDX handbook with linked pages, shared English/German catalogs, and an in-place locale switch
- document authoring, configuration, framework setup, limitations, and the native architecture boundary

## Architecture

FerroMark owns MDX recognition and its semantic event stream. `crates/palamedes` owns translation-unit selection, placeholder construction, message identity, diagnostics conversion, and JSX module lowering. The CLI and typed Node binding call that shared workflow; the Vite plugin only adapts configuration, diagnostics, and native code/maps to the host.

This keeps extraction and rendered compilation on one implementation instead of maintaining separate Markdown heuristics.

## Validation

- `RUSTUP_TOOLCHAIN=1.97.1 pnpm agent:check`
- `pnpm --filter @palamedes/site build`
- `node ./scripts/verify-examples-all.mjs --id vite-mdx`
- real Vite React and Solid integration builds in the Vite plugin test suite
- browser coverage for all three linked MDX pages and the German locale switch

Closes #477


swernerx:
## Review — request changes

> Posted as a comment because GitHub does not allow a formal review on your own PR.

Checked out at `2a2c620`, ran the Rust suite, and verified the load-bearing assumptions with probe builds and probe tests rather than reading alone.

The architecture call is right: one semantic pipeline shared by extraction and compilation instead of two Markdown heuristics. The cache stamp with the `CACHE_SCHEMA` bump is clean, and source-ranged diagnostics are threaded through properly.

But the feature does not currently run. Both documented setups fail the build, and there are four correctness bugs — two of which silently drop translations.

---

## 1. Blocker — nothing transforms the generated JSX

`palamedes:mdx` returns JSX under a `.mdx` id. Neither Rolldown, nor `@vitejs/plugin-react`, nor `vite-plugin-solid` touches `.mdx` — the extension is in no loader map.

I reproduced the exact setup from `docs/mdx.md:19` with a stub plugin (Vite 8.1.3, plugin-react 6.0.3):

```
[PARSE_ERROR] Unexpected JSX expression
   ╭─[ src/guide.mdx:3:5 ]
   │ Help: JSX syntax is disabled and should be enabled via the parser options
```

Same failure with `vite-plugin-solid`. Note that `react({ include: /\.(mdx|js|jsx|ts|tsx)$/ })` is **not** sufficient on Vite 8 — that governs Fast Refresh, not parsing.

What actually works (verified, build green):

- **React:** `build.rollupOptions.moduleTypes = { ".mdx": "jsx" }`. The plugin can set this itself from a `config()` hook, so users need zero extra configuration.
- **Solid:** `solid({ extensions: [".mdx"] })` — and `moduleTypes` must **not** be set here, otherwise Rolldown compiles the JSX with the React automatic runtime before `babel-preset-solid` ever runs.

So the fix is framework-dependent:

1. A `config()` hook that sets `moduleTypes` only for `framework: "react"`.
2. Solid needs an explicit documented step (`extensions: [".mdx"]`); the plugin cannot do this on the user's behalf.

Also, `packages/vite-plugin/package.json` declares `vite: ^3 || ^4 || ^5 || ^6 || ^7 || ^8`. On Vite ≤6 (esbuild) the knob is different again (`esbuild.include` / `optimizeDeps.esbuildOptions.loader`). Either cover that or narrow MDX support to Vite 7+.

**Root cause of the blocker: there is no end-to-end verification anywhere.** Every Vite test mocks `analyzeMdxNative` away (`packages/vite-plugin/src/index.test.ts:15`), and the Rust tests only assert the emitted string parses as TSX. No test ever renders a compiled MDX module. An `examples/*` app with one `.mdx` page wired into `pnpm verify:examples` would have caught this immediately — CONTRIBUTING already asks for exactly that when a change touches framework integration.

---

## 2. Correctness bugs

All confirmed with probe tests against the real compiler, not inferred.

### 2a. Prose that mentions the directive string disappears silently

`crates/palamedes/src/mdx.rs:780` checks not just the preceding comment but the unit's **text content**:

```rust
|| plain_text(&unit.roots).contains(&self.options.ignore_directive)
```

Input: `Use the palamedes-ignore comment to opt out.`
Result: `messages: []` — no warning, no diagnostic.

Any documentation page explaining the feature — including ours — loses its translation. This line should go; `is_ignored_unit` alone is the correct check.

### 2b. `palamedes-ignore` does not actually ignore

`finish_unit` decides on ignore only at the **end** of the unit, but images and JSX attributes push their messages during inline processing (`mdx.rs:753`, `mdx.rs:870`).

```mdx
{/* palamedes-ignore */}

See ![Diagram label](./d.png) here.
```

produces `messages: ["Diagram label"]`, and the generated module still contains:

```jsx
<img src="./d.png" alt={__palamedesGetI18n()._("2wNh4w8mgbU", …)} />
```

Identical behavior for `<Card title="Open settings" />`. The opt-out is not an opt-out. Either resolve ignore state before the inline phase, or roll back records produced during the unit in `finish_unit` (remember `self.messages.len()` at unit start).

### 2c. Generated TSX is invalid as soon as a URL contains a quote

`js_string()` emits JSON escapes that are then placed into **JSX string attributes** (`mdx.rs:608`, `:616`, `:638`, `:766`, `:997`). JSX attributes have no backslash escapes — `\"` terminates the string.

```
Read [the guide](<https://x.test/a"b>).
```

produces `<a href="https://x.test/a\"b" />`, and oxc reports `Unexpected token`. The build dies with no diagnostic pointing at the cause.

Fix: use expression containers consistently — `href={"…"}` rather than `href="…"`. Affects `alt=`, `src=` and `align=` too.

### 2d. HTML comments become visible page content and translatable messages

```mdx
<!-- Translator: keep short -->

Save now.
```

produces `messages: ["<!-- Translator: keep short -->", "Save now."]` and, in the module:

```jsx
<p><__PalamedesTrans message={"<!-- Translator: keep short -->"} /></p>
```

The comment renders as a paragraph. Same for `<!-- palamedes-ignore -->`, where the marker itself becomes visible text.

Follow-on consequence: `preceding_comment` (`mdx.rs:1260`) only supports `<!-- -->` and therefore effectively never fires — the `comment` field stays empty and translator comments do not work at all. `is_ignored_unit` accepts `<!--`, `{/*` and `//`; `preceding_comment` accepts only `<!--`. Only `{/* */}` behaves correctly (verified).

Suggestion: make `{/* */}` the canonical form, point `preceding_comment` at it, and either handle `<!-- -->` properly as a comment event or reject it with a diagnostic.

### 2e. Footnotes are silently mangled

```mdx
Text with a note[^1].

[^1]: The note body.
```

produces the message `Text with a note<0>^1</0>.` with `components={{ 0: <a href="The" /> }}` — the footnote definition is read as a link reference definition and the `href` is the first word of the body. `InlineEvent::FootnoteRef` is discarded at `mdx.rs:668`.

At minimum, document footnotes as unsupported; better, emit a diagnostic.

---

## 3. Other findings

**Front-matter lists are stringified.** `tags: [a, b]` becomes `export const frontmatter = {"tags":"[a, b]"}`. The hand-rolled scanner at `mdx.rs:1281` only understands `key: value`. Since `frontmatter` is an exported public API, silently corrupting a list into a string is bad — skip non-scalars, or parse YAML properly. Separately, the `BTreeMap` sorts keys alphabetically, so document order is lost.

**Identical expressions get duplicate placeholders.** `Hello {user.name} and again {user.name}` yields `"Hello {name} and again {name_1}"` with `values={{ name: (user.name), name_1: (user.name) }}`. Confusing for translators, and the message ID shifts with every added repetition. Dedupe on expression text in `RichUnit::value` (`mdx.rs:923`).

**No `components` prop / no MDXProvider.** `MDXContent(props = {})` (`mdx.rs:311`) never uses `props`, and `<p>`/`<h1>` are hardcoded. That is a real departure from the MDX convention and probably the first thing people will hit. Document it at minimum. The unused parameter will also trip `noUnusedParameters` in strict projects.

**Code block info string is dropped.** `CodeBlockStart { .. }` (`mdx.rs:432`) discards the language, so `<pre><code>` has no `className="language-ts"` and syntax highlighting is impossible. Fairly central for a documentation feature.

**Raw HTML keeps `class` instead of `className`**, producing a React runtime warning.

**Two different `runtimeModule` options.** `packages/vite-plugin/src/index.ts:83` (`@palamedes/runtime`, PO loader) and `mdx.runtimeModule` (`@palamedes/react/runtime`). Same name, different meaning, different default. Worth renaming while nothing is released.

**`matchesTransformFilter` is an extension-rewrite hack.** `index.ts:132` tests `.mdx` against the filter by rewriting the id to `.tsx`. If someone configures a glob (`include: "src/**/*.ts"`), MDX silently turns off. Prefer handling all `.mdx` when `mdx !== false`, optionally with a dedicated `mdx.include`.

**Catalog watching has no effect.** `index.ts:224-231` registers every `catalog × locale` file and `handleHotUpdate` invalidates MDX modules on `.po` changes. But the generated module does not depend on catalogs at all — translations resolve at render time through the runtime. The cache is keyed on `source + signature`, so the re-transform returns the identical result. Either drop this, or — better — actually consume `result.compiledIds` and apply `failOnMissing` to MDX. As it stands `compiledIds` is dead API surface in the plugin.

**`mdxCache` grows unbounded** and holds the source plus `sourcesContent` twice over. It is useless in build mode (each module transforms once) and duplicates Vite's own transform cache in dev. `mdxModuleIds` is never pruned when files are deleted.

**Rust defaults duplicated in TS.** `index.ts:207` re-specifies `framework: "react"` and `translatableAttributes: ["alt"]`, but not `ignoreDirective`. Inconsistent and drift-prone — pass `undefined` through and let Rust default.

**`translatableAttributes` replaces rather than extends.** Configuring `[title]` silently drops `alt`. Classic trap; belongs in the docs table.

**`ExtractCache::load` / `load_with_mdx_stamp`.** At `crates/palamedes/src/extract_cache.rs:142`, no production code calls `load` anymore, yet it stays public with a silent `""` stamp. And `reset_if_request_differs` is now `#[cfg(test)]`-only — a shim kept alive purely for old tests. A `CacheIdentity` struct parameter would keep this at one function.

**The glob widening is an unrelated behavior change.** `crates/palamedes-cli/src/commands/extract/sources.rs:69` goes from `{js,jsx,ts,tsx}` to `{js,jsx,mjs,cjs,ts,tsx,mts,cts,mdx}`. `.mjs/.cjs/.mts/.cts` have nothing to do with MDX and may start picking up build artifacts in existing projects. Should be called out in the PR body/changelog or split into its own commit.

**`extractMessages` takes no options.** `packages/extractor/src/index.ts` accepts `.mdx`, but `extract_messages_with_scopes` hardcodes `MdxOptions::default()` (`crates/palamedes/src/extract.rs:1338`). Front-matter fields and configured attributes are unreachable through that path, so the package advertises MDX support it only half has.

**`extract_mdx_messages` joins diagnostics with `", "`** (`mdx.rs:207`), producing one unreadable line for multiple errors; `\n` would fit better. Also, `source_range` builds a fresh `SourceLocator` over the whole file per diagnostic.

**`DefinitionDescriptionStart`** (`mdx.rs:485`) opens no unit, so inline events are silently discarded at `mdx.rs:576`. I could not reach the branch with the current FerroMark setup (definition lists parse as a paragraph), but once the feature is active, `<dd>` content vanishes from both output and extraction. Either open a unit or remove the branch.

---

## 4. Documentation

- `docs/mdx.md:19` — the setup snippet does not build (see §1). Both framework paths need the corrected, verified setup step.
- Nothing covers what happens when `@mdx-js/rollup` is already in the project. Both claim `.mdx`; Palamedes wins via `enforce: "pre"`. That should be explicit.
- Undocumented limits users will hit: no `components` prop, no language class on code blocks, footnotes broken, `context` is always `None` (so homonyms cannot be disambiguated), `translatableAttributes` replaces the default, `href`/`src` are hardcoded exclusions.
- `docs/mdx.md` is missing from `site/app/routes.ts` and `site/app/routes/docs/`, so the page will not appear on palamedes.dev. `llms.txt` / `llms-full.txt` are also unchanged.
- CONTRIBUTING asks for "the failure mode or diagnostics users will see" — the diagnostics section only covers the happy path for structurally broken MDX.

---

## 5. Missing test coverage

The five Rust tests cover the happy path. Untested and broken in each case: the ignore leak through images/attributes (§2b), the directive string in prose (§2a), quotes in URLs and alt text (§2c), HTML comments (§2d), footnotes (§2e), front-matter lists, and duplicate expressions.

`assert_valid_tsx` should run against **every** compile output in the test module, not just the React/Solid test — that alone would have surfaced §2c.

And there should be at least one `examples/*` app with an `.mdx` page wired into `pnpm verify:examples`.

---

Suggested order of work: §1 blocker → §2c (invalid TSX) → §2a/§2b (ignore semantics) → §2d.

swernerx:
## Re-review after `765bb3d` — one regression left, otherwise good

Rebuilt the native binding on this branch and re-ran the same probe suite plus the real builds, rather than reading the diff alone.

**The blocker is verified gone.** `config()` sets `moduleTypes` for React, `moduleType: "jsx"` comes back from the transform, Solid is deliberately excluded, and the Vite <7 guard gives a clear message. `mdx.integration.test.ts` builds real React *and* Solid entries through the full framework pipeline with no mocks — 13/13 green — and `examples/vite-mdx` is wired into `verify:examples`. That is exactly the missing layer.

All five correctness bugs are fixed. Same probes as before:

| | before | now |
|---|---|---|
| §2a prose containing the directive string | `messages: []` | extracted correctly |
| §2b ignore leak via image/attribute | `["Diagram label"]` + i18n call | `[]`, literal in output |
| §2c quote in URL | `!!! INVALID TSX` | valid, `href={"…"}` throughout |
| §2d HTML comment | own message + visible `<p>` | `comment: Some("Translator: keep short")` |
| §2e footnote | `<a href="The" />` | `UnsupportedFootnote`, no code emitted |

Also confirmed fixed from §3: frontmatter collections skipped instead of stringified, expressions deduplicated, `language-*` on code blocks, `className` instead of `class`, the `props` parameter dropped, the extension-rewrite filter hack replaced, `mdxCache` and catalog watching removed, `mdxModuleIds` pruned on delete, TS-side defaults no longer duplicated, the `.mjs/.cjs/.mts/.cts` glob widening reverted, `DefinitionDescriptionStart` now opens a unit, diagnostics joined with `\n`. Docs now cover the Solid setup, the Vite 7 requirement, the `@mdx-js/rollup` conflict and the known limits, and the site route plus `llms.txt` are in.

---

### One regression introduced by the fix commit

The `config()` hook loads the Palamedes config **eagerly**. Without a `palamedes.yaml`, the entire Vite startup now fails in `resolveConfig`:

```
Error: Could not find a Palamedes config. Expected one of palamedes.yaml, …
 ❯ getConfigLazy src/index.ts:113:16
 ❯ BasicMinimalPluginContext.config src/index.ts:192:21
 ❯ runConfigHook … vite/dist/node/chunks/node.js:36207:15
```

Checked against `main`: the same project builds fine there, so this is new in `765bb3d`, and it hits projects that use no MDX at all.

Suggested fix: swallow the config error inside `config()` and simply skip `moduleTypes` when no config is found. The `transform` hook already fails cleanly with a proper message once an actual `.mdx` module shows up, so nothing is lost.

---

### Still open — non-blocking follow-ups

- `runtimeModule` still exists twice with different meanings (plugin option vs. `mdx.runtimeModule`).
- `compiledIds` remains unused on the MDX path, and `failOnMissing` does not apply to MDX.
- `ExtractCache::load` plus the `#[cfg(test)]` shim for `reset_if_request_differs` are still there.
- `crates/palamedes/src/extract.rs:1343` still hardcodes `MdxOptions::default()`, so `@palamedes/extractor` cannot reach frontmatter fields or configured attributes.

LGTM once the eager-config regression is addressed. The rest is follow-up material.

swernerx:
Thanks for both detailed review rounds — all actionable findings are now addressed.

The original Vite blocker, correctness issues, documentation gaps, and missing end-to-end coverage were fixed in `765bb3d`. The configless-startup regression identified in the re-review was fixed separately in `5a0c585`.

The remaining non-blocking follow-ups are covered by `6855f5e`:

- The top-level `runtimeModule` is now the shared fallback for macro transforms and generated MDX modules. `mdx.runtimeModule` remains the explicit, more-specific MDX override.
- MDX `compiledIds` are now used by `failOnMissing` to validate every relevant target catalog. This is covered by a real Vite build with actual PO catalogs.
- Cache identity now flows through `ExtractCatalogMessagesOptions`; the `#[cfg(test)]` reset shim and raw MDX-stamp API are gone. I kept the pre-existing public `ExtractCache::load` method as a backwards-compatible default-options wrapper because that API already exists on `main`.
- The public Rust, Node, and `@palamedes/extractor` surfaces now accept MDX options. `createExtractor({ mdx })` supports configured attributes and frontmatter fields as well.

Validation is green: the full local `pnpm agent:check`, the documentation build, the `vite-mdx` smoke/browser verification, and all seven GitHub CI jobs pass. The PR is now ready for review.

PR #496, opened 2026-08-01
perf(runtime): generate executable catalog messages

swernerx:
## What changed

- generate one catalog map whose values are either constant strings or executable message functions
- lower Ferrocat's ICU AST to a small renderer-independent function ABI used directly by Core, React, and Solid
- use one canonical native catalog-module generator; the public TypeScript helper delegates to it instead of parsing and generating the same program again
- hoist plural/select branch functions and branch tables to module scope, so rendering does not allocate them
- keep hand-written string catalogs and invalid/unsupported generated patterns on the existing bounded lazy-parser path
- keep `renderMessage()` optional so older and custom i18n instances retain the previous React/Solid node-rendering path
- document the representation, direct `@palamedes/core` runtime dependency, and trade-offs in ADR-022 and the adapter guides

## Why

The first version of this PR attached a second, parallel AST map to the original string map. That removed first-render parsing, but duplicated dynamic keys and retained AST interpretation in the browser hot path.

The generated catalog is now the runtime program itself: static text stays data, while every dynamic translation is a compact function. There is no right-hand metadata structure and no AST traversal for first-party generated messages.

Ferrocat is also the only build-time ICU parser for generated catalogs. First-party loaders and the low-level TypeScript compatibility API now share the same Rust lowering and JavaScript generator, so subtle plural/select/quoting behavior cannot drift between two implementations.

## Concrete generated examples

The examples below use readable IDs and names; the real output uses compact internal IDs and minified local names.

### Static text

Catalog translation:

```icu
Welcome back
```

Generated value:

```js
welcome: "Welcome back"
```

This remains a plain string and is returned directly.

### Static text with a variable

Catalog translation:

```icu
Hello {name}
```

Generated value:

```js
greeting: (values, runtime) =>
  runtime.join("Hello ", runtime.value(values, "name"))
```

For `{ name: "Ada" }`, Core returns `"Hello Ada"`; React and Solid produce their native node values from the same function. No ICU parser or message-node interpreter runs.

### Plural

Catalog translation:

```icu
{count, plural, one {# message} other {# messages}}
```

Generated module code, expanded for readability:

```js
const one = (values, runtime, pluralValue) =>
  runtime.join(runtime.pound(pluralValue), " message")
const other = (values, runtime, pluralValue) =>
  runtime.join(runtime.pound(pluralValue), " messages")
const inboxBranches = { one, other }

const inbox = (values, runtime) =>
  runtime.plural(values, "count", 0, "plural", inboxBranches)
```

The branch functions and `inboxBranches` are created once when the module loads, not on every render. `{ count: 2 }` selects `other` and produces `"2 messages"`.

### Rich text

Catalog translation:

```icu
Read <strong>{title}</strong>
```

Generated value:

```js
rich: (values, runtime) =>
  runtime.join(
    "Read ",
    runtime.tag("strong", runtime.value(values, "title")),
  )
```

The React and Solid renderers turn `tag()` into the supplied framework component; Core's string renderer keeps the children.

The complete catalog is therefore one object:

```js
export const messages = defineCompiledCatalog({
  welcome: "Welcome back",
  greeting,
  inbox,
  rich,
})
```

## Compatibility and fallback behavior

- `CatalogMessages` remains the public `Record<string, string>` type for hand-written catalogs.
- Generated catalogs use the branded `CompiledCatalogMessages` type with `string | CompiledMessage` values.
- Spreading a generated catalog drops the constant-string optimization safely; function entries remain executable.
- JSON is intentionally not a transport for generated executable catalogs because JSON omits functions. Source PO/FCL data and hand-written string catalogs remain serializable.
- Invalid or unsupported generated patterns emit a function that calls the existing lazy parser, preserving `onError` and source fallback behavior.
- `getMessage()` reconstructs a canonical ICU pattern for a generated function. `getMessageNodes()` remains a compatibility API and parses that reconstructed pattern only when explicitly called; normal rendering through a current Core instance does neither.
- React and Solid use `renderMessage()` when it exists and fall back to `getMessageNodes()` for older or custom instances.
- Generated modules import `defineCompiledCatalog()` from `@palamedes/core`; adapter and custom-integration guides now require Core as a direct runtime dependency for strict package-manager layouts.

## Runtime and bundle impact

- A local warm-path microbenchmark for one variable message over 1,000,000 renders ran the generated function about 1.8–1.9× faster than the cached string/AST path. This is an illustrative local diagnostic, not a committed benchmark baseline.
- The Vite MDX production example changes from 211.31 kB / 67.33 kB gzip on `main` to 214.37 kB / 68.19 kB gzip: +3.06 kB raw / +0.86 kB gzip.
- Compared with the former parallel-AST version of this PR, the raw example bundle drops from 215.40 kB to 214.37 kB while removing the parallel catalog structure and AST interpretation hot path.

## Validation

  - full package build and test matrix
  - published TypeScript declarations and native type parity
  - TypeScript typecheck
  - Oxlint, ESLint, and formatting
  - release-set validation
  - Rust formatting, workspace Clippy with `-D warnings`, and all workspace tests
- executable generated-module parity tests for variables, formatters, ICU quoting, rich text, selects, plurals, offsets, ordinals, and nested select/plural pound semantics
- compatibility tests for React and Solid instances without `renderMessage()`
- direct native-generator test plus a TypeScript-wrapper delegation test
- Vite MDX production build and emitted-bundle inspection


swernerx (commented):
## Review

Strong PR overall. The executable-catalog design is clean: one branded map, constants stay strings, dynamic messages become renderer-neutral functions over a small runtime ABI, and branch tables are hoisted to module scope. Error/fallback parity with the old `renderResilient` path is preserved by moving the try/catch into core's `renderMessage()`, and the test coverage is genuinely good (executable-module import test via `data:` URL, parity assertions against `formatMessagePattern`, nested select/plural pound semantics, offsets, ordinals). ADR-022 and the docs updates accurately describe the trade-offs. Also a nice touch: the computed object keys (`["id"]:`) plus `Object.create(null)` / `Object.hasOwn` in `load()` close the `__proto__`-message-id hazard the old `JSON.stringify` module had.

Three points I'd like to see addressed or consciously accepted before merge:

### 1. `@palamedes/core` may not resolve from generated catalog modules (strict pnpm)

The generated module now begins with `import{defineCompiledCatalog}from"@palamedes/core"`, and catalog modules are the app's own `.po` files transformed in place (`PO_FILE_REGEX` in `packages/vite-plugin/src/index.ts`). That import is therefore resolved from the **app's** `node_modules`. The React README still says `pnpm add @palamedes/react` only — with pnpm's strict layout, `@palamedes/core` is a transitive dependency and is *not* resolvable from app source, so the import fails at build/dev time. All examples happen to depend on `@palamedes/core` directly, which is why the Vite MDX validation didn't catch it. Suggested fixes: resolve the specifier in the plugins (e.g. `resolveId` to the plugin-resolved absolute path), or document core as a required direct dependency in the adapter/plugin READMEs.

### 2. `renderMessage` is a hard new requirement on `PalamedesI18n`

The interface gains a required `renderMessage` member and the React/Solid adapters call it unconditionally. The previous code explicitly guarded `i18n.reportError?.()` because "instances reach the adapters through the untyped global runtime bridge and may predate this hook" — that exact skew (new adapter + older core instance on the bridge) now throws `TypeError: i18n.renderMessage is not a function` instead of degrading. Custom `PalamedesI18n` implementations (which the old docs explicitly sanctioned) also break at compile time. If the packages are strictly released in lockstep this may be acceptable, but then it should be called out as a breaking change — the CHANGELOG currently lists this only under *Performance Improvements*. Alternatively: make `renderMessage` optional and keep the old `getMessageNodes()`-walk as the adapter fallback when it's absent.

### 3. Two parallel code generators with no cross-parity test

The lowering now exists twice: Rust (`crates/palamedes-node/src/catalog.rs`, fed by Ferrocat's AST) and TypeScript (`packages/transform/src/catalogLoader.ts`, fed by `parseMessagePattern`), both live in production paths. The semantics currently line up (including the subtle `#`-in-select-inside-plural and quoted-literal cases), but each generator is only tested against its own parser. A shared fixture catalog asserting both pipelines emit the same module source — or at least render identical output through the Core string runtime — would keep them from drifting apart silently.

### Minor

- `formatMessageNodes()` in `messageFormat.ts` is a private one-caller wrapper around `renderNodesToString` — could be inlined.
- No dedup of identical branch functions across messages (e.g. many `{count, plural, one {#} other {#}}` patterns each emit their own branches). Fine for now; worth remembering if catalog bundle size becomes a topic.

None of this touches the core design, which I think is right. Points 1 and 2 are the ones I'd resolve before merging.

---

swernerx (commented):
## Follow-up review of 1004360

All three concerns from my previous review are resolved — nicely, in fact:

1. **`@palamedes/core` resolution** — documented as a required direct dependency in the React and Remix READMEs (Solid already listed it), with the rationale spelled out, and the constraint is now recorded in ADR-022 for custom integrations. Choosing documentation over plugin-side `resolveId` is a reasonable call.

2. **`renderMessage` compatibility** — the interface member is optional again, and both adapters route through `renderI18nMessage()`, which falls back to the previous `getMessageNodes()` walk with full error/fallback parity when the hook is absent. The new legacy-instance tests in React and Solid cover exactly the skew scenario I was worried about, and the CHANGELOG now describes the behavior.

3. **Dual generators** — better than what I asked for: instead of a parity test, the TypeScript generator is gone entirely. The native Ferrocat-backed renderer is the single source of truth, exposed as `renderCatalogModule` over the NAPI boundary, with `@palamedes/transform` delegating to it (plus a delegation-equality test). `@palamedes/core` correctly moved to transform's devDependencies, and ADR-022 documents the single-generator decision.

The `formatMessageNodes` indirection was inlined too. The only observable side effect of the centralization is that module entries are now emitted in sorted key order (BTreeMap) rather than insertion order — harmless, and the updated loader test reflects it.

LGTM — no remaining change requests from my side.

---

PR #499, opened 2026-08-02
perf(runtime): keep generated browser bundles parser-free

swernerx:
## What changed

Generated production code now uses explicit parser-free entrypoints:

- catalog modules import `defineCompiledCatalog` from `@palamedes/core/compiled`
- transformed React and Solid `Trans` calls import the framework `/compiled` entrypoint
- generated MDX uses those same framework entrypoints
- apps that load only generated catalogs can create their instance through `@palamedes/core/compiled`
- the package roots remain the compatibility path for hand-written ICU strings, parser APIs, and direct runtime component patterns

Core's shared state machine and Intl formatting helpers were split from the ICU parser, so the optimized entrypoint has no static path back to parsing code. React keeps its hook-free `react-server` condition on the new entrypoint.

## Concrete generated output

A static translation remains a plain string:

```js
// "Hallo"
["greeting"]: "Hallo"
```

A translation with a named variable becomes a direct function:

```js
// "Hallo {name}"
const message = (values, runtime) =>
  runtime.join("Hallo ", runtime.value(values, "name"))
```

A plural becomes one function plus module-hoisted branches:

```js
// "{count, plural, one {# Nachricht} other {# Nachrichten}}"
const one = (values, runtime, pluralValue) =>
  runtime.join(runtime.pound(pluralValue), " Nachricht")
const other = (values, runtime, pluralValue) =>
  runtime.join(runtime.pound(pluralValue), " Nachrichten")
const branches = { one, other }
const message = (values, runtime) =>
  runtime.plural(values, "count", 0, "plural", branches)
```

All three values live in the same branded catalog map; there is no parallel string/AST structure. The browser executes valid dynamic messages directly.

## Runtime boundary

Generated-only applications use:

```ts
import {
  createI18n,
  type CompiledCatalogMessages,
} from "@palamedes/core/compiled"
```

Applications that intentionally load hand-written ICU strings continue to use:

```ts
import { createI18n } from "@palamedes/core"
```

The parser-free factory narrows `load()` to `CompiledCatalogMessages` and also rejects unbranded string catalogs at runtime. Invalid or unsupported generated patterns report through `onError` and degrade to their raw fallback instead of silently pulling the parser back into the bundle.

## Browser result

Measured with the real Vite MDX production example:

| JavaScript | Before | After | Delta |
| --- | ---: | ---: | ---: |
| raw | 214.65 kB | 209.29 kB | -5.36 kB (-2.5%) |
| gzip (Vite) | 68.28 kB | 66.68 kB | -1.60 kB (-2.3%) |

The compatibility build proves the stable ICU parser sentinel is live, while the resulting browser chunk contains none of it. `pnpm benchmark:runtime-browser` reproduces the build and reports raw, gzip, and Brotli sizes.

## Validation

  - package builds
  - formatting and lint
  - all JS/TS tests and type checks
  - published declaration checks
  - Rustfmt, Clippy, and workspace tests
- Vite MDX production build plus `check-runtime-browser-bundle.mjs`
- representative React Router, SolidStart, TanStack, and Waku production builds
- Next example TypeScript check
- CI now runs the real parser-free browser bundle check on full Linux jobs

ADR-023 records the entrypoint boundary and compatibility policy.


swernerx (commented):
## Review

This is a well-executed follow-up to #496. The central design move — injecting `I18nPatternSupport` into a shared `createI18nRuntime()` so the root entry closes over the parser while `/compiled` simply omits it — is the right shape: no conditional imports, no environment sniffing, and the boundary is statically verifiable, which the new CI check then actually verifies. Things I specifically like:

- The `runtimeFormat.ts` / `runtimeChoice.ts` extraction is a faithful move — I compared the extracted functions against `main` and the only change is the standalone `MessageFormat` union replacing `MessageFormattedArgumentNode["format"]`, which is exactly what's needed to cut the type-level dependency on the parser AST.
- The parser-free factory failing fast at `load()` for unbranded catalogs, and `getMessageNodes()` throwing with a message that names the fix, are good developer-experience choices. The `Symbol.for` catalog brand from #496 pays off here: both entrypoints agree on branding even when the bundler duplicates `compiledMessage` between the two entry chunks.
- The react-server condition on `./compiled` plus the exports test, the ImportCollector accepting both root and `/compiled` sources for dedup, and ADR-023's honest rejection of "just rely on tree-shaking" are all solid.
- Lazy-pattern degradation is coherent end to end: the generated `r.pattern()` call throws in the parser-free string runtime, core catches it, reports through `onError`, and returns the raw fallback — covered by `compiled.test.ts`.

Questions and suggestions — none of these are blocking:

### 1. Only one of 25 examples exercises the parser-free path end to end

`vite-mdx` is the only example importing `createI18n` from `@palamedes/core/compiled`; the other 24 keep the root factory and only tighten their catalog types to `CompiledCatalogMessages`. That means the CI parser-free guard covers exactly one bundler/framework combination, and the "representative React Router / SolidStart / TanStack / Waku production builds" in the validation section still ship the parser. If those apps load only generated catalogs, migrating at least one example per framework would widen the guard considerably; if they intentionally stay on the compatibility path (e.g. because they demo runtime `<Plural>`/`<Select>` components), a sentence in ADR-023 or the examples README saying so would prevent the obvious "why doesn't the example use the optimized entry?" confusion.

### 2. The bundle check is string-brittle

`check-runtime-browser-bundle.mjs` greps four literal parser diagnostic strings. If someone rewords those error messages later, the check goes green while the parser could be back in the bundle — a silent false pass on the exact regression it exists to catch. Two cheap hardenings: assert against a dedicated sentinel constant that lives in the parser module and is guaranteed to survive minification (e.g. part of an error message declared once), and/or add a negative control that runs the same signature scan against a root-entry build and asserts it *does* match, so the signatures themselves are validated.

### 3. `pattern()` in the compiled adapters re-enters lookup with the pattern as id

`compiledShared.tsx` / `solid/compiled.tsx` implement the `pattern` runtime operation as `i18n.getMessageNodes(pattern, { message: pattern, reportMissing: false })`. On a full-runtime instance, if a loaded catalog contains an entry whose id equals the raw pattern text — the norm for hand-written natural-language-id catalogs — this resolves that catalog entry instead of parsing the given pattern. That's arguably even desirable (it's a translation lookup), but it's a behavioral difference from the root adapters' direct `parseMessagePattern(pattern)` and looks accidental. A short comment stating the intent (or routing through a parse-only path when one is available) would settle it.

### 4. Minor fallback parity gap in the legacy path

The compiled adapters' `renderI18nMessage` catch goes straight to `runtime.join(fallback)`, while the root adapters first retry `runtime.pattern(fallback, values)`. With a legacy (parser-capable) instance, a broken catalog message therefore degrades to raw source text with visible `{braces}` under the compiled Trans, but to a formatted source message under the root Trans. Probably irrelevant in practice — legacy instance + compiled component is already a corner — but it's a one-line difference worth either aligning or acknowledging.

### 5. Adapter duplication, again but smaller

`compiledShared.tsx` and `solid/compiled.tsx` are near-copies of the node-walking halves of `shared.tsx` / `index.tsx` within the same packages. After #496's generator consolidation this is the same pattern one level up. It's contained (same package, shared `resolveChoice`/format helpers from core), so fine to ship as-is — but if the root Trans could be expressed as the compiled Trans plus injected parser support, mirroring what core did with `createI18nRuntime`, both packages would lose ~150 duplicated lines each. Candidate for a follow-up rather than this PR.

Overall: correctness looks right everywhere I probed (I found no bugs), the measured win is honest and now guarded in CI, and the compatibility story is explicit. Points 1 and 2 are the ones I'd act on — both are small.

---

PR #508, opened 2026-08-03
feat(cli): resolve native binary at runtime

### LINKED ISSUE #506: feat(cli): eliminate the postinstall lifecycle script
## Problem

`@palamedes/cli` currently declares a `postinstall` script. The script does **not** fetch code from the network: it resolves the already-installed platform-specific optional dependency, copies its native `pmds` binary into `@palamedes/cli/bin`, and sets executable permissions.

That is still awkward for consumers with strict supply-chain policies such as pnpm `strictDepBuilds`: they must grant `@palamedes/cli` lifecycle-script execution via `allowBuilds`, even though the only intended operation is selecting the locked platform binary. This grants more install-time capability than Palamedes actually needs.

The concrete adoption case is 3DegreesGroupInc/EACManagement#1138, where exact Palamedes versions and lockfile integrity are already used, but the lifecycle-script exception remains a review concern.

## Proposed solution

Make the CLI usable without any install lifecycle script. One possible shape:

- keep publishing the platform-specific optional packages;
- make the JavaScript `bin/pmds` launcher resolve the package for the current OS/architecture/libc at runtime;
- spawn that package's binary directly instead of copying it during `postinstall`;
- preserve exit codes, signals, stdio, and useful errors for missing optional dependencies;
- remove the `postinstall` entry from `@palamedes/cli`.

Acceptance coverage should install/run the published package with lifecycle scripts disabled (for example, `pnpm install --ignore-scripts`) and verify `pmds --version` on supported targets.

## Affected surface

CLI

## Alternatives considered

- Keep the current consumer-side `allowBuilds` entry. This works but permanently broadens install-time privileges.
- Vendor native binaries in each consuming repository. This duplicates release artifacts, bloats repositories, complicates multi-platform development/CI, and loses the normal npm integrity workflow.
- Download a binary from `postinstall`. Palamedes correctly does not do this today, and it would make the supply-chain concern worse.

## Compatibility notes

This should not change CLI arguments or commands. The launcher needs to preserve native process behavior, particularly exit status, signals, stdio, Windows execution, and Linux libc selection.

swernerx:
## What changed

- Resolve the matching platform package and native `pmds` executable when the CLI runs.
- Keep built-in and explicit plugin dispatch on the existing Node wrapper, including native resolution errors for plugin-hosted built-ins.
- Drop the `postinstall` copy/chmod flow and its installer.
- Add a packed-package smoke fixture that runs without lifecycle hooks and covers `pmds version` plus `pmds --version`.
- Run the lifecycle-free CLI package test on macOS and Windows CI as well as the existing Linux test jobs.

## Why

The CLI only needs to select an already-installed optional platform package. Doing that at runtime removes the install-time privilege requirement without changing command arguments, dispatch, output, exit codes, signals, or stdio behavior.

Closes #506

## Checks

- `RUSTUP_TOOLCHAIN=stable pnpm --filter @palamedes/cli test`
- `pnpm format:check`
- `pnpm lint`
- `pnpm --filter @palamedes/cli exec tsc --noEmit -p tsconfig.json`
- `pnpm check:release-set`
- `cargo fmt --all --check`
- `git diff --check`

## Follow-ups

None.


swernerx:
Superseded by #512, which includes the lifecycle-free runtime resolution from this PR and moves binary plugin hosting into Rust. Closing this alternative in favor of #512.

PR #512, opened 2026-08-03
feat(cli): host binary plugins in Rust

### LINKED ISSUE #506: feat(cli): eliminate the postinstall lifecycle script

swernerx:
## What changed

- Keep the lifecycle-free runtime platform resolution from #508.
- Move external command parsing, configured plugin resolution, protocol handling, diagnostics, JSON/text output, and exit-code propagation into the native Rust `pmds` binary.
- Restrict CLI plugins to standalone executables. Remove the ESM/CJS plugin host, `@palamedes/cli/plugin`, and their JavaScript fixtures.
- Restrict native CLI configuration to `palamedes.yaml`, `.yml`, `.json`, and `.toml`; JavaScript and TypeScript configs remain outside the Rust CLI boundary.
- Resolve either a directly configured binary package/path or a platform-neutral plugin package whose installed optional dependency declares a matching `palamedes.pluginBinary`.
- Exercise the packed, lifecycle-free npm package against a real Rust plugin binary and update the ADRs/API documentation.

## Why

Binary-only plugins no longer need a JavaScript host. Keeping dispatch in Node would leave Palamedes with two CLI runtimes and duplicate configuration, routing, output, and protocol semantics.

The small Node launcher remains only as a packaging bridge: package managers such as pnpm do not expose a transitive optional dependency's `bin` as the public `pmds` executable of the platform-neutral package. It selects the installed platform package and immediately spawns Rust; it does not inspect commands or configuration.

This is an alternative to #508 and includes its lifecycle-free runtime resolution.

Closes #506

## Compatibility

This intentionally removes the unreleased JavaScript CLI plugin surface and executable JS/TS CLI configuration. Framework integrations and `@palamedes/config` may continue to read their legacy config formats; the native CLI does not.

## Checks

- `RUSTUP_TOOLCHAIN=stable pnpm build`
- `RUSTUP_TOOLCHAIN=stable pnpm test`
- `RUSTUP_TOOLCHAIN=stable pnpm check-types`
- `pnpm format:check`
- `pnpm lint`
- `pnpm check:release-set`
- `rustup run stable cargo clippy --workspace --all-targets --locked -- -D warnings`
- `rustup run stable cargo test --workspace --locked`
- `rustup run stable cargo fmt --all --check`
- `git diff --check`


crates/palamedes-cli/src/plugins.rs
```diff
@@ -0,0 +1,1195 @@
...
+                    format!(
+                        "Could not write a request to binary plugin \"{}\": {error}",
+                        resolved.specifier
+                    ),
+                ));
+            }
+        }
+    }
+
+    let mut stdout = String::new();
```
swernerx: The host writes the whole request, then reads stdout to EOF, which has two consequences:

1. **No streaming**: `output` events render only after the plugin exits. `CommandContext::output`'s doc ("Streams a line of human-readable output through the host") oversells this — a long-running plugin command appears silent on stdout until completion (only stderr passes through live).
2. **Deadlock hazard**: an SDK plugin reads stdin to EOF before writing, but a non-compliant plugin that emits more than a pipe buffer (~64 KB) of events before draining stdin will deadlock against a request that also exceeds the buffer (the serialized config can get there on large projects): the host blocks in `write_all`, the plugin blocks writing stdout, and the failure mode is a silent hang instead of a protocol error.

Reading stdout incrementally while the child runs (line-by-line, or the write on a thread) fixes both, and would let non-JSON mode print `output` events as they arrive.

crates/palamedes-cli/src/plugins.rs
```diff
@@ -0,0 +1,1195 @@
...
+        PluginFailure::new(
+            "PLUGIN_BINARY_PROTOCOL",
+            format!(
+                "Could not serialize a request for binary plugin \"{}\": {error}",
+                resolved.specifier
+            ),
+        )
+    })?;
+    request_bytes.push(b'\n');
+    let mut child = Command::new(&resolved.binary_path)
```
swernerx: There's no `SIGINT`/`SIGTERM` forwarding to the child — ADR 018 now scopes cancellation to "members of the invoking terminal's process group", which covers Ctrl+C but not a signal sent to `pmds` directly (`timeout --signal=TERM`, CI supervisors, `kill <pid>`). In that case the host dies and the plugin child is orphaned and keeps running. The removed JS host explicitly forwarded both signals to the child.

Worth either forwarding (a small `signal_hook`-style handler that kills the child) or calling the limitation out explicitly in `docs/api/cli-binary-plugin.md` so plugin authors know a supervisor kill won't reach them.

crates/palamedes-cli/src/plugins.rs
```diff
@@ -0,0 +1,1195 @@
...
+    finish_run(&plugin.resolved, result)
+}
+
+fn load_registry(
+    config: &LoadedConfig,
+    cwd: &Path,
+    native_executable: &Path,
+) -> Result<BTreeMap<String, LoadedPlugin>, PluginFailure> {
+    let mut registry = BTreeMap::new();
+    for declaration in &config.plugins {
```
swernerx, line 186: Two consequences of building the full registry per invocation:

1. **Fail-fast across unrelated plugins**: any resolution or describe failure aborts the whole registry, so a missing platform package for plugin B makes `pmds a …` fail too. Consider degrading failures of *non-requested* namespaces to warning diagnostics and failing only when the requested namespace can't be served (or genuinely collides).
2. **Describe-all cost**: every `pmds <plugin> …` spawns every configured plugin binary before dispatch. Fine at 1–2 plugins; if the list grows, a manifest cache keyed on binary mtime would keep this flat.

Both are defensible as-is (deterministic, and collision detection needs all manifests) — flagging so the tradeoff is a decision rather than an accident, ideally noted in ADR 018.

crates/palamedes-cli/src/plugins.rs
```diff
@@ -0,0 +1,1195 @@
...
+
+fn resolved_executable(
+    specifier: &str,
+    binary_path: PathBuf,
+) -> Result<ResolvedPlugin, PluginFailure> {
+    if matches!(
+        binary_path
+            .extension()
+            .and_then(|extension| extension.to_str()),
+        Some("js" | "mjs" | "cjs" | "ts")
```
swernerx: The denylist misses `mts` and `cts` (and arguably `jsx`/`tsx`): a `pluginBinary: "./plugin.mts"` passes this guard and fails later at spawn with a less actionable error (`PLUGIN_BINARY_SPAWN_FAILED` — or on a Unix system with a shebang line it may even run). Extending the list keeps the nice `PLUGIN_BINARY_REQUIRED` message for all script flavors.

crates/palamedes-plugin/src/protocol.rs
```diff
@@ -19,7 +19,7 @@ pub struct Request {
     /// Protocol major the host speaks; must match [`PROTOCOL_VERSION`].
     #[serde(rename = "palamedesBinaryPluginProtocol")]
     pub protocol_version: u64,
-    /// Version of the `@palamedes/cli` host package.
+    /// Version of the native `pmds` host.
```
swernerx, line 22: Same rename missed one spot: `CommandContext::host_version` in `lib.rs` (line 272) still says "Version of the `@palamedes/cli` host package." — `hostVersion` is now the native CLI's `CARGO_PKG_VERSION`.

swernerx (commented):
## Review

Solid direction and a clean cut: one parser and one dispatch table in Rust, the npm launcher reduced to platform selection, and the ADRs/docs updated to match reality. I verified `cargo test -p palamedes-cli -p palamedes-plugin --locked` (54 tests) and `cargo clippy --all-targets -- -D warnings` locally on this branch — both green. Catalog payload parity with the removed JS host checks out (pseudo-locale excluded, `{locale}` substitution, absolute paths), and the packed-tarball smoke test asserting no `postinstall` and no copied sidecar is a nice regression guard.

### Main findings (inline comments)

1. **Plugin stdout is fully buffered** — `output` events only render after the plugin exits, and a non-SDK plugin can deadlock against a large request (details inline at `invoke_binary`).
2. **No signal forwarding to plugin children** — a direct `SIGTERM`/`SIGINT` to `pmds` (CI supervisors, `timeout`) orphans the running plugin; the removed JS host forwarded both.
3. **Registry is describe-all + fail-fast** — every plugin invocation spawns every configured plugin, and one broken plugin blocks all others' commands.
4. Small items: `mts`/`cts` slip past the script-extension rejection; one stale `@palamedes/cli` doc reference on `CommandContext::host_version`.

### Test coverage

The deleted `plugin-host.test.mjs`/`binary-plugin.test.mjs` suites (~580 lines) covered host-side behavior that now has no direct Rust equivalent: `validate_manifest` (protocol mismatch, built-in collision, invalid names), the `--json` envelope shape on success *and* failure, and namespace collision across plugins. The new unit tests cover parsing/resolution/`finish_run`, and `smoke.mjs` covers one happy path E2E. A few host-side unit tests for `validate_manifest` and `emit_result`/`emit_failure`, plus optionally a Rust integration test that spawns the `inspect` example binary, would restore most of that.

### Minor UX notes (no action required)

- With `external_subcommand`, a typo like `pmds extractt` no longer gets clap's "did you mean `extract`" — it reports `PLUGIN_NAMESPACE_UNKNOWN`; in a directory without a config it even reports `PLUGIN_CONFIG_FAILED`. A similarity hint against built-ins would soften this.
- A malformed `plugins` entry (e.g. `- [name]` with a missing options value) surfaces serde's opaque "did not match any variant of untagged enum" message; a custom `Deserialize` could produce a friendlier config error later.
- pnpm's symlink layout for meta-plugin resolution is covered by reasoning + canonicalization but not by any automated test (the smoke test builds a plain `node_modules` tree by hand).

None of these block the PR from my side — 1–3 are worth deciding deliberately (fix or document) before the protocol leaves Preview.

swernerx:
Follow-up commits addressing the review findings:

- **4bece8b** resolves the three main inline findings: `invoke_binary` now writes the request from a helper thread and streams stdout line by line (output events render live in text mode, and the write/read deadlock window is gone); the host forwards `SIGINT`/`SIGTERM` to the running plugin child (verified end-to-end: `kill -TERM <pmds>` → plugin exits, host returns 143, no orphans); and a plugin that fails to resolve/describe now blocks only its own namespace — other plugin commands run with a `PLUGIN_UNAVAILABLE` warning. Also: `mts`/`cts`/`jsx`/`tsx` join the script rejection, the stale `@palamedes/cli` doc on `host_version` is fixed, and host-side tests cover manifest validation, script rejection, streaming, and registry degradation. ADR-018 and the binary-plugin docs record the new semantics.
- **2553af4** adds the platform-package `pmds` bins (Node-free path for known-platform environments) plus launcher signal forwarding, with a packed-tarball smoke guard.
- **3149fba** records the launcher distribution decision as ADR-024.

Known limitation (unchanged by design): after a forwarded signal the host reads stdout to EOF, so a plugin's orphaned grandchild holding the pipe delays host exit — same semantics as a shell command substitution; SDK plugins are single-process and unaffected.

Checks run locally on this branch: `cargo fmt --check`, `cargo clippy --all-targets --locked -D warnings` (both crates), `cargo test --locked` (57 tests), `pnpm format:check`, `pnpm lint`, CLI platform tests, and the packed smoke test.

Issue #516, opened 2026-08-03
Spike an Oxlint/ESLint adapter for native Palamedes source diagnostics


swernerx:
## Parent

Follow-up to #510. The native Palamedes analysis remains the source of truth.

## Goal

Determine whether a small ESLint-compatible JavaScript plugin can present native Palamedes source diagnostics through Oxlint and ESLint without reimplementing macro recognition or rule semantics.

The retired plugin already demonstrated that non-type-aware JavaScript rules can execute through Oxlint. This spike evaluates the stronger adapter design: call the native Palamedes analysis once per relevant file and map its structured diagnostics into standard lint rule UX.

## Questions to answer

1. Can an Oxlint JS plugin reliably load the published Palamedes native binding on every supported platform?
2. Can all enabled Palamedes rule facades share one native analysis per file?
3. How should UTF-8 byte ranges from Rust map to ESLint/Oxlint source locations and UTF-16 JavaScript indices?
4. Can per-rule severities/options and inline disable directives work without duplicating Palamedes configuration?
5. Can suggestions or fixes be transported safely?
6. What editor/LSP experience does Oxlint provide for the adapter?
7. What is the cold/warm overhead of Oxlint parsing a file and Palamedes parsing relevant files again?
8. What happens when the native package is missing, version-incompatible, or loaded from a monorepo?

## Prototype scope

- Expose at least two non-type-aware native diagnostics as separate lint rule names.
- Skip files that cannot contain Palamedes source constructs before crossing into native analysis.
- Cache one analysis result per source version/file within the lint run.
- Verify aliased imports, React, Solid, and non-ASCII ranges.
- Compare standard ESLint-compatible `create` with Oxlint's `createOnce` API where useful.
- Record the behaviour of rule-level Oxlint suppressions versus native `pmds` suppressions.

## Deliverables / acceptance criteria

- [ ] Minimal runnable adapter against current Oxlint and ESLint versions.
- [ ] Correct source ranges for ASCII and non-ASCII fixtures.
- [ ] Evidence that multiple enabled rule facades do not repeat native analysis.
- [ ] Editor/LSP screenshots or a reproducible manual verification note.
- [ ] Cold/warm time and memory comparison against `pmds lint` and a pure JS control.
- [ ] Packaging/versioning failure modes documented.
- [ ] Recommendation: publish, wait for a named Oxlint capability, keep CLI-only, or use a different editor protocol.

## Non-goals

- Copying the native rules into JavaScript.
[...]

Issue #510, opened 2026-08-03
Define Palamedes source-authoring diagnostics from the retired Lingui plugin scope


swernerx:
## Context

`sebastian-software/eslint-plugin-lingui-typescript` is being retired because
we no longer build on Lingui and cannot sustainably maintain a Lingui-specific
ESLint package. Its nine rules are still useful as a requirements inventory,
but Palamedes should not inherit Lingui compatibility or copy the implementation.

Palamedes already owns parsing, transformation, extraction, and authoring errors
through its native OXC-based core. Source-message diagnostics should therefore
be designed around shared Palamedes source facts, with an explicit product
boundary:

- Palamedes OSS owns technical correctness, interoperability, and high-signal
  general authoring DX.
- Palamedes+ may add organization-specific policy, linguistic quality,
  proprietary knowledge, semantic analysis, and professional workflow features.

Whether a check is local or deterministic does not decide its product placement.
Palamedes OSS must remain independently useful, while commercial rule logic and
corpora should not first be implemented in the MIT repository.

Legacy reference:
https://github.com/sebastian-software/eslint-plugin-lingui-typescript

## Disposition of the legacy rules

| Legacy rule | Disposition |
| --- | --- |
| `no-nested-macros` | Already enforced as a fatal extraction/transform error through `NestedMessageMacro`. Keep this in the extractor. |
| `t-call-in-function` | Already enforced for eager translation macros through `TranslationMacroOutsideFunction`. Keep this in the extractor. |
| `no-expression-in-message` | Partially covered by `UnnamedPlaceholder`: expressions without a stable placeholder name fail. Do not copy the stricter Lingui rule blindly; Palamedes intentionally supports useful named member/accessor expressions. Define any further Core restriction from extraction or runtime semantics. |
| `no-single-variables-to-translate` | Implement as a high-signal Core diagnostic for messages with no authored translatable content; see #514. |
| `no-single-tag-to-translate` | Evaluate the narrower empty-component case as an opt-in Core diagnostic with React, Solid, and MDX counterexamples; see #514. |
| `text-restrictions` | Project and organization-specific source policy belongs to the optional Palamedes+ quality layer. Public proposal #513 is closed as not planned in OSS. |
| `consistent-plural-format` | Do not port as a Lingui compatibility rule. Palamedes choice macros canonicalize to ICU and use their own authoring model. Add a Core rule only if Palamedes supports two genuinely conflicting source forms. |
[...]

PR #520, opened 2026-08-03
Add native source-authoring diagnostics and lint workflow

### LINKED ISSUE #514: Diagnose Palamedes messages with no translatable content
## Parent

Follow-up to #510. Requires semantic source-message facts from the shared native analysis.

## Motivation

A message made only from runtime values or an empty component gives translators no language to translate and usually indicates that the localization boundary was placed incorrectly.

The legacy rule name `no-single-variables-to-translate` was narrower than its behaviour: JSX containing multiple value expressions and no literal text had the same problem. Palamedes should define the semantic condition directly instead of preserving the Lingui rule name.

### Placeholder-only message

Report a message whose meaningful authored content consists only of one or more value placeholders and contains no literal translatable text.

Examples to report:

```tsx
t`${status}`
<Trans>{firstName}{lastName}</Trans>
```

Examples not to report:

```tsx
t`Status: ${status}`
<Trans>{"{name}"}</Trans>
```

The implementation must use semantic source parts, not pattern matching on the normalized ICU string.

### Empty component-only message

Add a narrower opt-in diagnostic for a message whose only meaningful content is one empty component placeholder, for example:

```tsx
<Trans><Button /></Trans>
```

Do not report rich text that contains translatable text inside the component:

```tsx
<Trans><strong>Delete project</strong></Trans>
```

Fragments, comments, and formatting whitespace must not change classification.

## UX

- Use Palamedes-specific stable diagnostic codes and names that describe the actual semantic condition.
- Point to the complete message macro and explain how to move translation to surrounding authored text.
- Make placeholder-only a candidate recommended warning.
- Keep component-only opt-in until the React, Solid, and MDX corpus demonstrates sufficiently low false positives.

## Acceptance criteria

- [ ] Tagged templates, descriptor messages, `Trans`, aliases, React, and Solid have positive and negative fixtures.
- [ ] One and multiple value placeholders are handled consistently.
- [ ] Literal brace text is not mistaken for a placeholder.
- [ ] Empty component placeholders are distinguished from components wrapping translated text.
- [ ] JSX comments, fragments, entities, and whitespace normalization do not create false positives.
- [ ] MDX-valid rich-text cases are represented in the corpus before enabling the component rule by default.
- [ ] Diagnostics have exact ranges and actionable help.

## Non-goals

- Recreating Lingui compatibility.
- Detecting arbitrary unlocalized strings.
- Applying the rule to plural/select branches until branch-level semantics have a separate reviewed proposal.

### LINKED ISSUE #515: Add a shared native source-analysis result and diagnostic model
## Parent

Follow-up to #510.

## Problem

Palamedes already parses and classifies its source macros in the native core, but extraction and transformation currently keep separate parsing/import-collection paths. `ExtractedMessageRecord` also represents the normalized catalog message rather than the authored source shape and does not carry an exact source range.

New source-authoring diagnostics must not introduce a third parser, a third macro resolver, or rules that infer source syntax from the normalized ICU message after relevant facts have been discarded.

## Goal

Introduce one native analysis contract that can produce extracted messages and structured source diagnostics from the same OXC parse, macro classification, and semantic message decoding.

## Proposed shape

- Define a general `SourceRange` with inclusive start byte, exclusive end byte, and one-based line/column.
- Define `SourceDiagnostic` with stable code, severity, file, primary range, human-readable message, actionable help, and an optional related range.
- Define `SourceAnalysisResult` containing extracted messages and diagnostics.
- Add an `analyze_source` entry point for JS, TS, JSX, and TSX.
- Generalize or align the existing MDX ranged-diagnostic model rather than creating an incompatible second schema.
- Preserve semantic message facts until diagnostics have run: authored literal fragments, value placeholders, component placeholders, source surface, and the enclosing macro range.
- Factor the common Palamedes macro package/import/alias classification so extraction and transformation cannot drift.
- Keep current extraction and transform APIs as compatibility wrappers over the shared components where practical.
- Expose the structured result through the Node binding for future adapters and editor integrations.

## Error and diagnostic boundary

- Parse failures and non-extractable/runtime-incorrect authoring remain fatal.
- Core authoring findings are non-fatal diagnostics whose configured severity is applied by callers. Neutral facts and result transport may also be consumed by Palamedes+ without moving premium policy semantics into Core.
- A valid file may produce multiple quality diagnostics in deterministic source order.
- This issue should not silently change the current `pmds extract` exit behaviour.

## Acceptance criteria

- [ ] One OXC parse and one Palamedes macro classification serve extraction plus non-type-aware source diagnostics within an analysis invocation.
- [ ] Aliased imports and React/Solid macro packages resolve identically across extraction, transformation, and analysis fixtures.
- [ ] Diagnostics include exact stable ranges before normalized message construction loses source-shape information.
- [ ] Literal braces are distinguishable from value placeholders, and nested JSX structure is distinguishable from flattened component markup.
- [ ] JS, TS, JSX, TSX, and MDX results use a documented compatible diagnostic schema.
- [ ] Existing extraction and transform fixture behaviour remains unchanged.
- [ ] Native and Node-facing serialization is deterministic.
- [ ] The realistic extraction benchmark records the incremental cost of collecting semantic facts and diagnostics.

## Non-goals

- Implementing individual quality rules.
- Adding the CLI command.
- Building an Oxlint plugin.
- Adding advanced TypeScript type-aware or flow-based product intelligence; that is outside the Core contract and planned in the optional Palamedes+ layer.

### LINKED ISSUE #517: Suggest `<Trans>` for directly renderable JSX translations
## Parent

Follow-up to #510.

## Motivation

Palamedes deliberately supports both `t` and `<Trans>`, but equivalent code is often substantially easier to read in JSX when it uses the JSX-native authoring surface:

```tsx
<div>{t`Welcome ${name}`}</div>
```

versus:

```tsx
<div><Trans>Welcome {name}</Trans></div>
```

This is a DX/readability recommendation, not a claim that inline `t` is invalid. The rule was added in the retired standalone plugin as a local improvement and was not part of Lingui's original ESLint rule set.

## Proposed diagnostic

Suggest `<Trans>` when a Palamedes `t` message is directly renderable as a JSX child, including carefully defined conditional and logical branches.

Candidate inputs include tagged templates and static descriptor calls whose comment/context metadata can be preserved.

## Required exclusions

- JSX attribute values where a string is required.
- Function arguments, arrays, object properties, and other non-render positions.
- Text-only or structurally restricted parents where inserting a component is invalid or unsafe, including SVG `title`/`desc` and relevant HTML content models such as `option` or `textarea`.
- Cases where converting changes evaluation, reactivity, server/client, or hydration semantics.
- Non-Palamedes functions that happen to be named `t`.

## React and Solid

The rule must resolve aliased Palamedes imports and evaluate React and Solid separately. An automatic fix may add or reuse `Trans` only when the correct framework module is unambiguous and the import cannot capture an existing binding.

If framework inference or semantic equivalence is uncertain, emit a suggestion without a fix rather than guessing.

## Acceptance criteria

- [ ] Direct JSX child, conditional, and logical render positions have positive fixtures.
- [ ] Attributes, call arguments, arrays, object values, and restricted DOM/SVG parents have negative fixtures.
- [ ] Tagged-template expressions and supported static descriptor forms preserve text, expressions, comments, and context.
- [ ] Aliased imports and unrelated local `t` functions are classified correctly.
- [ ] React and Solid fixtures verify runtime/reactivity equivalence.
- [ ] Fixes are offered only when the framework import and rewrite are unambiguous.
- [ ] The final disposition records whether the rule belongs in the recommended warning set or remains opt-in.
- [ ] The diagnostic clearly says this is a readability suggestion and that `t` remains supported.

## Non-goals

- Banning `t` inside JSX.
- Requiring React-style authoring outside directly renderable JSX positions.
- Rewriting code when semantic equivalence cannot be demonstrated.

### LINKED ISSUE #518: Add `pmds lint` for native source-authoring diagnostics
## Parent

Follow-up to #510. Depends on the shared source-analysis contract.

## Goal

Add a non-mutating `pmds lint` command for Palamedes-specific source-authoring diagnostics. This is not a general JavaScript linter: it validates Palamedes messages and authoring surfaces while reusing the native analysis used by extraction.

## Command behaviour

- Discover files from the same `catalogs[].include` / `exclude` configuration as extraction.
- Deduplicate files matched by overlapping catalogs.
- Do not create or update catalogs.
- Keep deterministic file/range/code ordering regardless of worker scheduling.
- Render concise human-readable diagnostics with actionable help.
- Support a stable JSON result for CI, Palamedes+, and future editor adapters.
- Support configured severities and rule options without changing extraction correctness semantics.
- Support `--fail-on error|warning` with documented exit codes.
- Provide narrow code-specific suppressions such as next-line or exact-range suppression, and report malformed/unknown suppression codes.
- Reuse or extend the extraction cache so running `extract` and `lint` does not repeatedly parse unchanged files when the cached analysis is compatible.

## Configuration questions to settle

- Final top-level configuration name and kebab-case schema.
- Default/recommended versus opt-in rules.
- Severity overrides and per-rule options.
- Whether a CLI path override is needed beyond configured catalog sources.
- Cache invalidation when rule options change.
- Whether unused suppressions should be reportable.

## Acceptance criteria

- [ ] `pmds lint` works for JS, TS, JSX, TSX, and configured MDX sources.
- [ ] Human output includes file, line, column, severity, stable code, message, and help.
- [ ] `--json` emits one documented deterministic document.
- [ ] `--fail-on` never hides the diagnostics that explain the failure.
- [ ] Suppressions are code-specific and narrowly scoped.
- [ ] Overlapping catalogs do not duplicate diagnostics.
- [ ] Cached and uncached results are byte-for-byte equivalent.
- [ ] CI and configuration documentation includes minimal examples.

## Non-goals

- Replacing Oxlint or ESLint.
- Advanced type-aware unlocalized-string intelligence; that is outside the Core lint command and planned in the optional Palamedes+ layer.
- Writing source fixes in the first version.

swernerx:
## Summary

This implements the open-source source-authoring diagnostics discussed in #510. It:

- adds one native Rust analysis result for extracted messages and structured diagnostics
- preserves literal, value-placeholder, component-tree, source-surface, and exact macro-range facts until diagnostics run
- aligns MDX diagnostics with the shared source-range schema while preserving existing extract failures
- shares Palamedes macro import and alias classification between extraction and transformation
- exposes the analysis contract and configurable rule levels through `analyzeSourceNative`
- diagnoses placeholder-only messages by default and empty-component-only messages when opted in
- suggests `<Trans>` in conservative, directly renderable JSX positions while keeping `t` fully supported
- adds non-mutating `pmds lint` with human and deterministic JSON output, configured severities, `--fail-on`, and code-specific line suppressions
- reports malformed, unknown, and unused suppressions
- extends the extraction cache with compatible diagnostics so `extract` and `lint` reuse the same native parse
- records extract-versus-analysis cost in the normal and large proof benchmarks

## Increments

- [x] Complete the shared analysis, semantic source-fact, and benchmark work in #515
- [x] Diagnose messages without translatable content (#514)
- [x] Suggest `<Trans>` in safe, directly renderable JSX positions (#517)
- [x] Add the non-mutating `pmds lint` workflow (#518)

Each increment is preserved as a separate commit. The Oxlint/ESLint adapter work in #516 remains separate because it has a different packaging and compatibility risk.

## Verification

- `cargo test --workspace --locked` (Rust 1.97.1)
- `cargo clippy -p palamedes -p palamedes-cli -p palamedes-node --all-targets --locked -- -D warnings` (Rust 1.97.1)
- `pnpm build`
- `pnpm test`
- `pnpm check-types`
- `pnpm lint`
- `pnpm format:check`
- generated native type declarations checked against napi-rs output
- proof benchmark with normal and 10,000-message extract/source-analysis lanes
- cached and uncached lint JSON verified byte-for-byte equivalent

Closes #515
Closes #514
Closes #517
Closes #518

### DIFF
--- a/crates/palamedes-cli/src/commands/lint.rs
+++ b/crates/palamedes-cli/src/commands/lint.rs
@@ -1,25 +1,18 @@
 //! `pmds lint` — non-mutating Palamedes source-authoring diagnostics.
 
 use std::collections::BTreeSet;
-use std::fs;
 use std::path::{Path, PathBuf};
 
 use clap::{Args, ValueEnum};
 use palamedes::{
     analyze_source_file_cached, default_cache_path, ExtractCache, ExtractCatalogMessagesOptions,
-    SourceDiagnostic, SourceDiagnosticSeverity, SourceRange,
+    SourceDiagnostic, SourceDiagnosticSeverity, SourceRange, SOURCE_DIAGNOSTIC_CODES,
 };
 use serde::Serialize;
 
 use crate::command::{render_json, Command, Context};
 use crate::commands::extract::sources::{collect_source_files, sort_and_dedupe_paths};
 use crate::error::CliError;
-
-const KNOWN_DIAGNOSTIC_CODES: &[&str] = &[
-    "pmds/no-placeholder-only-message",
-    "pmds/no-empty-component-only-message",
-    "pmds/prefer-trans-in-jsx",
-];
 
 #[derive(Debug, Args)]
 pub struct LintOptions {
@@ -101,16 +94,6 @@ impl Command for LintOptions {
 
         for path in &files {
             let filename = display_source_path(path, &config.root_dir);
-            let source = match fs::read_to_string(path) {
-                Ok(source) => source,
-                Err(error) => {
-                    failed_files.push(LintFileFailure {
-                        file: filename,
-                        message: error.to_string(),
-                    });
-                    continue;
-                }
-            };
             match analyze_source_file_cached(
                 &path.to_string_lossy(),
                 &filename,
@@ -120,7 +103,7 @@ impl Command for LintOptions {
             ) {
                 Ok(result) => {
                     let (mut file_diagnostics, file_suppressed) =
-                        apply_suppressions(&source, &filename, result.diagnostics);
+                        apply_suppressions(&result.source, &filename, result.analysis.diagnostics);
                     diagnostics.append(&mut file_diagnostics);
                     suppressed += file_suppressed;
                 }
@@ -329,7 +312,7 @@ fn parse_suppressions(source: &str, filename: &str) -> (Vec<Suppression>, Vec<So
                 continue;
             }
             for code in codes {
-                if KNOWN_DIAGNOSTIC_CODES.contains(&code) {
+                if SOURCE_DIAGNOSTIC_CODES.contains(&code) {
                     suppressions.push(Suppression {
                         line: line_index + 1 + line_delta,
                         code: code.to_owned(),
--- a/crates/palamedes-cli/src/commands/lint.rs
+++ b/crates/palamedes-cli/src/commands/lint.rs
@@ -6,8 +6,8 @@ use std::path::{Path, PathBuf};
 
 use clap::{Args, ValueEnum};
 use palamedes::{
-    analyze_source_with_options, SourceAnalysisOptions, SourceDiagnostic, SourceDiagnosticSeverity,
-    SourceRange,
+    analyze_source_file_cached, default_cache_path, ExtractCache, ExtractCatalogMessagesOptions,
+    SourceDiagnostic, SourceDiagnosticSeverity, SourceRange,
 };
 use serde::Serialize;
 
@@ -32,6 +32,9 @@ pub struct LintOptions {
     /// Fail on error or warning diagnostics.
     #[arg(long, default_value = "error")]
     fail_on: LintFailOn,
+    /// Ignore and do not write the shared source-analysis cache.
+    #[arg(long)]
+    no_cache: bool,
 }
 
 #[derive(Clone, Copy, Debug, ValueEnum)]
@@ -40,7 +43,7 @@ enum LintFailOn {
     Warning,
 }
 
-#[derive(Debug, Serialize)]
+#[derive(Debug, PartialEq, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct LintResult {
     diagnostics: Vec<SourceDiagnostic>,
@@ -48,14 +51,14 @@ pub struct LintResult {
     summary: LintSummary,
 }
 
-#[derive(Debug, Serialize)]
+#[derive(Debug, PartialEq, Serialize)]
 #[serde(rename_all = "camelCase")]
 struct LintFileFailure {
     file: String,
     message: String,
 }
 
-#[derive(Debug, Default, Serialize)]
+#[derive(Debug, Default, PartialEq, Serialize)]
 #[serde(rename_all = "camelCase")]
 struct LintSummary {
     files: usize,
@@ -80,9 +83,20 @@ impl Command for LintOptions {
         let mut diagnostics = Vec::new();
         let mut failed_files = Vec::new();
         let mut suppressed = 0usize;
-        let analysis_options = SourceAnalysisOptions {
+        let analysis_options = ExtractCatalogMessagesOptions {
+            reference_scopes: config.reference_scopes,
             mdx: config.mdx.clone(),
             rules: config.lint.rules.clone().into(),
+        };
+        let cache_path = default_cache_path(&config.root_dir);
+        let mut cache = if self.no_cache || !config.extract_cache {
+            ExtractCache::disabled()
+        } else {
+            ExtractCache::load_with_options(
+                &cache_path,
+                &config.source_reference_root.to_string_lossy(),
+                &analysis_options,
+            )
         };
 
         for path in &files {
@@ -97,7 +111,13 @@ impl Command for LintOptions {
                     continue;
                 }
             };
-            match analyze_source_with_options(&source, &filename, &analysis_options) {
+            match analyze_source_file_cached(
+                &path.to_string_lossy(),
+                &filename,
+                &config.source_reference_root.to_string_lossy(),
+                &analysis_options,
+                &mut cache,
+            ) {
                 Ok(result) => {
                     let (mut file_diagnostics, file_suppressed) =
                         apply_suppressions(&source, &filename, result.diagnostics);
@@ -110,6 +130,13 @@ impl Command for LintOptions {
                 }),
             }
         }
+
+        let cache_keys = files
+            .iter()
+            .map(|path| path.to_string_lossy().into_owned())
+            .collect::<Vec<_>>();
+        cache.retain_paths(&cache_keys.iter().map(String::as_str).collect());
+        let _ = cache.save(&cache_path);
 
         diagnostics.sort_by(|left, right| {
             (
@@ -222,6 +249,7 @@ fn severity_name(severity: SourceDiagnosticSeverity) -> &'static str {
 struct Suppression {
     line: usize,
     code: String,
+    primary: SourceRange,
 }
 
 fn apply_suppressions(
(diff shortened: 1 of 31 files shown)


crates/palamedes-cli/src/commands/lint.rs
```diff
@@ -0,0 +1,551 @@
...
+            ExtractCache::disabled()
+        } else {
+            ExtractCache::load_with_options(
+                &cache_path,
+                &config.source_reference_root.to_string_lossy(),
+                &analysis_options,
+            )
+        };
+
+        for path in &files {
```
swernerx, line 95: Two cold-run observations on this loop, both fine to defer. Every cache miss reads the file twice: once here for suppressions and again inside `analyze_source_file_cached`. Besides the extra IO, an edit landing between the two reads would pair suppression lines from the newer content with diagnostics from the older one; letting `analyze_source_file_cached` accept or return the source text would close both. Separately, extraction pushes the same analysis through rayon while lint runs serial, so a cold `pmds lint` on a large tree will trail `pmds extract --no-cache` noticeably. Probably irrelevant as long as the shared cache keeps lint warm, just worth remembering if big-repo timings come up.

crates/palamedes-cli/src/commands/lint.rs
```diff
@@ -0,0 +1,551 @@
...
+    analyze_source_file_cached, default_cache_path, ExtractCache, ExtractCatalogMessagesOptions,
+    SourceDiagnostic, SourceDiagnosticSeverity, SourceRange,
+};
+use serde::Serialize;
+
+use crate::command::{render_json, Command, Context};
+use crate::commands::extract::sources::{collect_source_files, sort_and_dedupe_paths};
+use crate::error::CliError;
+
+const KNOWN_DIAGNOSTIC_CODES: &[&str] = &[
```
swernerx: These string-duplicate the codes emitted in `crates/palamedes/src/extract.rs`. When a fourth rule lands in the core crate, suppressing it reports `pmds/unknown-suppression-code` here until someone remembers this constant. Exporting the known codes from `palamedes` next to `SourceRuleOptions` would make that drift impossible. Optional.

crates/palamedes/src/extract.rs
```diff
@@ -342,15 +461,13 @@ impl<'a> Visit<'a> for ExtractionVisitor<'a> {
         }
 
         let Some(tag_name) = it.opening_element.name.get_identifier_name() else {
-            walk::walk_jsx_element(self, it);
+            self.walk_jsx_element_children(it, None);
             return;
         };
+        let tag_name = tag_name.as_str().to_owned();
```
swernerx: Hot-path note, take it or leave it: every JSX element now allocates a `String` for the parent stack, and `visit_jsx_child` collects render spans even when all three rules are `off` (the `extract_messages` compat path). Given how deliberately this file avoids per-node work elsewhere, a `Vec<&'a str>` stack (`Atom` derefs to `&'a str`) plus an early-out when no rule is enabled would keep the disabled path at its old cost. The new benchmark lane suggests the current delta is acceptable, so this is purely optional.

swernerx (commented):
Reviewed the full diff against #510/#514/#515/#517/#518. Nothing blocking; a few optional notes inline.

What holds up well under scrutiny: one parse genuinely serves both extraction and diagnostics, classification runs on the authored part tree instead of the normalized ICU string, and macro import/alias resolution is now shared with the transform side via `source_macros.rs`, so those two can't drift anymore. The byte-equivalence claims are actually tested in both directions (extract-warms-lint and cached-vs-cold JSON). The `SourceLocator` consolidation also quietly fixes the old byte-based columns to the documented Unicode-scalar contract.

Things I checked specifically and found covered:

- placeholder-only / empty-component-only against the #514 acceptance list: literal brace text, fragments, JSX comments, aliases, Solid all have fixtures
- prefer-trans exclusions against #517: attributes, call args, arrays, object values, restricted DOM/SVG parents; `||`/`??` are left alone, only conditional branches and the right side of `&&` opt in, unrelated local `t` is ignored
- suppressions: unknown, malformed, and unused are all reported, and codes are rule-specific as #518 asked
- cache: the stamp now covers rule levels, schema bumped 2 to 3, failed files are still never cached, and the Node batch-extract path stays on a disabled cache, so there is no stamp thrash between Node-driven extraction and CLI lint
- YAML `off` stays a string through both the Rust and the TS loader, with tests on both sides

crates/palamedes/src/extract.rs
swernerx: Addressed in e11e7b9. The JSX parent stack now stores arena-backed `&str` references, and `visit_jsx_child` skips render-span collection entirely when `prefer-trans-in-jsx` is disabled.

crates/palamedes-cli/src/commands/lint.rs
swernerx: Addressed in e11e7b9. Core now exports the built-in diagnostic constants and `SOURCE_DIAGNOSTIC_CODES`; rule emission and CLI suppression validation both consume that single source of truth.

crates/palamedes-cli/src/commands/lint.rs
swernerx: Addressed in e11e7b9. `analyze_source_file_cached` now owns the single read and returns the exact source together with its analysis. A cache hit is validated across that read, so suppressions and diagnostics always use the same text. I left cold-run parallelization for a separate measured follow-up because the shared mutable cache makes that a distinct change.

PR #522, opened 2026-08-04
fix(solid): type compiled runtime renderer

agent-werner:
## What this fixes

The Palamedes Deploy Site workflow is currently failing during the Ardo/TypeDoc prebuild step because `packages/solid/src/transShared.tsx` is compiled under strict TypeScript settings and the shared compiled-runtime renderer leaves several callback parameters implicit.

This adds the missing host-renderer parameter types and gives `renderNode` an explicit fallback return so TypeDoc can convert the Solid package API again.

## Validation

- `corepack pnpm --filter @palamedes/solid build`
- `corepack pnpm --filter @palamedes/core build`
- `corepack pnpm --filter @palamedes/runtime build`
- `corepack pnpm build:site`
- `corepack pnpm --filter @palamedes/solid exec tsc --noEmit -p tsconfig.json`
- `corepack pnpm format:check -- packages/solid/src/transShared.tsx`

Note: full `corepack pnpm check-types` still fails locally before this package on the existing Darwin checkout because `@palamedes/core-node` native package types are not resolvable for `packages/extractor`; the targeted Solid typecheck and the failing site build path both pass.

### DIFF
--- a/packages/solid/src/transShared.tsx
+++ b/packages/solid/src/transShared.tsx
@@ -94,32 +94,32 @@ export function createSolidMessageRuntime(
   const runtime: CompiledMessageRuntime<JSX.Element[]> = createCompiledMessageRuntime<
     JSX.Element[]
   >(locale, {
-    pattern(pattern, values) {
+    pattern(pattern: string, values: Record<string, unknown>) {
       const nodes = parsePattern(i18n, pattern, fallbackParser)
       return renderNodes(nodes, values, runtime, locale)
     },
-    join(...parts) {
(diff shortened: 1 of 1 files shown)


swernerx (approved):
Reviewed: Die expliziten Renderer-Typen entsprechen dem `ExecutableMessageRenderer<JSX.Element[]>`-Vertrag, der Fallback ist für die aktuell exhaustive `MessageNode`-Union verhaltensneutral, und alle CI-Jobs einschließlich Site-Build und Typecheck sind grün. ✅

PR #539, opened 2026-08-05
fix(next): preserve i18n scope across RSC suspension

### LINKED ISSUE #538: fix(runtime): preserve server i18n scope across Next RSC suspension
## Summary

`createServerI18nScope().activate()` can lose the active i18n instance during a Next.js App Router production render when React suspends and later resumes server prerendering of a Client Component subtree. A compiled macro then calls `getI18n()` without an active store and the request fails with:

```
Error: No active server i18n instance. Configure @palamedes/runtime
with setServerI18nGetter() before translated code runs.
```

This is the suspend/resume variant left after #134 was closed by #142. The helper introduced there still activates both stores with `AsyncLocalStorage.enterWith()`.

## Environment

- `@palamedes/runtime` 1.12.0
- Next.js 16 App Router
- React Server Components / server prerendering of Client Components
- Node runtime
- Production build/deployment

The published `dist/server.mjs` implementation is byte-identical between Palamedes 1.11.0 and 1.12.0, so the 1.12 client catalog-boundary changes do not affect this server path.

## Reproduction shape

Create one module-level server scope and activate a fresh request-local instance synchronously before the initializer's first await:

```ts
const serverI18n = createServerI18nScope<PalamedesI18n>()

export function initPageI18n() {
  const state = getRequestI18nState()
  serverI18n.activate(state.i18n)
  return state.catalogPromise
}
```

Then render translated Client Components through a Next App Router layout/page. When a subtree suspends and resumes during the production server pass, a later compiled macro can execute in an async context where the shared store is empty:

```ts
getI18n()._(...)
```

Observed behavior:

- `next dev` renders normally.
- A production build/deployment returns 500 on otherwise public routes.
- The failure can be intermittent depending on the rendered/suspended subtree.
- Restoring a server getter fallback removes the production 500s without changing the 1.12 client boundary.

This was confirmed while validating 3DegreesGroupInc/EACManagement#1673. Before the fallback, affected production-preview routes failed consistently or intermittently; after the fallback, the same Web and Admin routes rendered successfully across repeated requests.

## Current implementation

```ts
activate(i18n) {
  sharedState.active.enterWith(i18n)
  storage.enterWith(i18n)
  return i18n
}
```

`enterWith()` binds the instance to the current async execution context. React may resume rendering in another context, where `sharedState.active.getStore()` returns `undefined`.

## Application workaround

The application currently overrides the server getter to prefer Palamedes' active shared scope and otherwise return a fixed default-locale source catalog:

```ts
setServerI18nGetter(() => {
  const state = globalThis[
    Symbol.for("palamedes.runtime.serverI18nScopeState")
  ]
  return state?.active.getStore() ?? fixedDefaultLocaleI18n
})
```

The fallback is deliberately fixed to the default locale rather than the last request's instance, preventing cross-request locale leakage. It avoids a 500 but cannot preserve the correct non-default server locale after the scope has already been lost, so it should not be the steady-state solution.

## Expected

Palamedes should provide a server integration whose request-local instance remains available for the complete React server render, including suspension and resumption, without requiring an application fallback.

## Suggested acceptance criteria

- [ ] A real Next.js production-build integration test suspends and resumes a translated Client Component subtree during server rendering.
- [ ] `getI18n()` still resolves the active request instance after resumption.
- [ ] Concurrent requests with different locales remain isolated.
- [ ] No global "last active request" fallback is required.
- [ ] The supported Next/RSC activation pattern and its lifetime are documented.

## Related

- #134
- #142
- #423 (different dual-module-graph failure mode with the same error text)

swernerx:
## Summary

Next can resume Client Component server rendering in an async context captured before `scope.activate()`. The plain `AsyncLocalStorage` scope is missing there, which made translated Client Components fail during production prerendering.

This PR:

- adds `createNextServerI18nScope()` under `@palamedes/next-plugin/server`, keyed to Next's stable App Router render store
- keeps render instances in a `WeakMap`, with no process-global "last request" fallback
- preserves nested `run()` overrides and sibling async-branch isolation
- moves the Next examples to the adapter and documents its Node/Next 16 lifecycle and compatibility boundary
- adds a real production-build regression that suspends a Client Component and checks concurrent English/German requests

## Issue

Closes #538.

## Validation

- `pnpm --filter @palamedes/runtime test` (28 tests)
- `pnpm --filter @palamedes/next-plugin test` (19 tests)
- focused Runtime and Next adapter TypeScript checks and builds
- `pnpm format:check`
- `pnpm lint`
- `pnpm --filter @palamedes/example-nextjs-cookie test:rsc-scope` (optimized Next 16.2 production build plus 12 concurrent alternating `en`/`de` requests)

## Follow-up

None planned. Next doesn't expose a public request identity that spans both App Router server-render passes, so the adapter uses Next 16's internal render storage. The compatibility boundary and build-time failure mode are documented.

### DIFF
--- a/examples/nextjs-cookie/package.json
+++ b/examples/nextjs-cookie/package.json
@@ -6,11 +6,13 @@
     "dev": "next dev --port 4010",
     "build": "next build && node ../../scripts/normalize-next-required-server-files.mjs",
     "start": "next start --port 4010",
-    "extract": "pmds extract"
+    "extract": "pmds extract",
+    "test:rsc-scope": "next build --webpack && node ./scripts/rsc-scope-production.test.mjs"
   },
   "dependencies": {
     "@palamedes/core": "workspace:^",
     "@palamedes/example-ui": "workspace:^",
+    "@palamedes/next-plugin": "workspace:^",
     "@palamedes/react": "workspace:^",
     "@palamedes/runtime": "workspace:^",
     "next": "^16.2.12",
@@ -21,7 +23,6 @@
   "devDependencies": {
     "@palamedes/cli": "workspace:^",
     "@palamedes/config": "workspace:^",
-    "@palamedes/next-plugin": "workspace:^",
     "@types/node": "^24.0.0",
     "@types/react": "^19.2.18",
     "@types/react-dom": "^19.2.4",
--- a/examples/nextjs-cookie/package.json
+++ b/examples/nextjs-cookie/package.json
@@ -7,7 +7,8 @@
     "build": "next build && node ../../scripts/normalize-next-required-server-files.mjs",
     "start": "next start --port 4010",
     "extract": "pmds extract",
-    "test:rsc-scope": "next build --webpack && node ./scripts/rsc-scope-production.test.mjs"
+    "test:rsc-scope": "next build && node ./scripts/rsc-scope-production.test.mjs",
+    "test:rsc-scope:webpack": "next build --webpack && node ./scripts/rsc-scope-production.test.mjs"
   },
   "dependencies": {
     "@palamedes/core": "workspace:^",
--- a/packages/runtime/src/server.ts
+++ b/packages/runtime/src/server.ts
@@ -1,13 +1,24 @@
 import { AsyncLocalStorage } from "node:async_hooks"
 
-import { type I18nInstance, type ServerI18nScope, setServerI18nGetter } from "./index"
+import {
+  type CreateServerI18nScopeOptions,
+  type I18nInstance,
+  type ServerI18nScope,
+  setServerI18nGetter,
+} from "./index"
 
-export type { ServerI18nScope } from "./index"
+export type { CreateServerI18nScopeOptions, ServerI18nScope } from "./index"
 
 const SERVER_SCOPE_STATE_KEY = Symbol.for("palamedes.runtime.serverI18nScopeState")
 
 type ServerScopeState = {
   active: AsyncLocalStorage<I18nInstance>
+  activate(i18n: I18nInstance): void
+  get(): I18nInstance | undefined
+  getRun(): I18nInstance | undefined
+  requestI18n: WeakMap<object, I18nInstance>
+  requestKeyProviders: Map<symbol, () => object | undefined>
+  run<Result>(i18n: I18nInstance, callback: () => Result): Result
 }
 
 function getServerScopeState(): ServerScopeState {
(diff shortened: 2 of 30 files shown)


examples/nextjs-cookie/package.json
```diff
@@ -6,11 +6,13 @@
     "dev": "next dev --port 4010",
     "build": "next build && node ../../scripts/normalize-next-required-server-files.mjs",
     "start": "next start --port 4010",
-    "extract": "pmds extract"
+    "extract": "pmds extract",
+    "test:rsc-scope": "next build --webpack && node ./scripts/rsc-scope-production.test.mjs"
```
swernerx: This builds with `--webpack`, but Next 16 defaults to Turbopack and the example's own `build` script uses that default. Since `verify:examples` isn't wired into CI, this webpack run is the only Next production build CI executes — and it exercises the non-default bundler. The whole fix rests on bundler-specific behavior (`.external.js` modules resolving to one shared singleton across the vendored and node_modules graphs); that should behave identically under Turbopack, but "should" is the category #538 came from. Either point the probe at the default bundler (or run both — the step is already isolated to a single matrix cell), or leave a one-line comment here saying why webpack is the deliberate choice.

packages/runtime/src/server.ts
```diff
@@ -17,35 +28,76 @@ function getServerScopeState(): ServerScopeState {
     return existing
```
swernerx: This returns an existing state without checking its shape. The state is shared via `Symbol.for`, which is deliberately cross-copy — but after this PR it's also cross-version: if an older runtime copy (1.12 or earlier) created the state first, it has no `activate`/`get`/`run`, and the new `activateServerI18n()` path in index.ts then throws a hard TypeError. That foreign code touches this symbol in practice is documented by #423 and the application workaround quoted in #538. A feature-detect here that upgrades an old-shaped state in place (reusing its existing `active` storage) would be cheap insurance. Not blocking — it takes two different installed runtime versions in one server process to hit.

swernerx (commented):
Detailed review pass, recorded as comments so the findings can be picked up, extended, or corrected. Verdict up front: I'd merge this. The layering is exactly right, and there's only one point I want resolved or explained before merge — the bundler choice in the regression test (inline comment on the example's package.json). Two further inline comments are optional hardenings.

What stands out positively:

- The runtime stays framework-neutral. `requestKeyProvider` is a small generic hook, and everything Next-specific lives in the 25-line adapter behind `@palamedes/next-plugin/server`. That's the right definition of lean: little code in the right place, not little code at any cost.
- Keying on `work-async-storage.external.js` is the technically correct choice. The `.external` modules are exactly the files Next itself shares as singletons between its vendored server bundle and node_modules (the same mechanism `headers()`/`cookies()` rely on), and the WorkStore is the only identity that spans both App Router server render passes including suspension. Making a removed internal fail at application build time instead of silently falling back to stale or cross-request state is the right call, and the README documents it as such, with an honest compatibility statement (peer `^16`, verified against 16.2).
- No leak surface: instances hang off the render object in a WeakMap, provider registration is bounded by the stable symbol id, and the HMR replacement path is actually tested (the stale provider provably stops being called).
- The test pyramid covers every acceptance criterion from #538: five targeted runtime unit tests (context resumption via `AsyncLocalStorage.snapshot()`, nested `run()` over the request-key fallback, sibling-branch isolation, concurrent request keys, HMR provider replacement) plus the dual-module-graph test with `vi.resetModules()`, an adapter test against the real `workAsyncStorage` rather than a mock, and a production E2E that genuinely suspends a Client Component and fires 12 concurrent alternating en/de requests. The probe design is thought through — server-side WeakMap resource for the delay, resolved promise in the browser so hydration stays clean — and the server teardown in the script is robust.
- Docs are updated consistently across all touchpoints (three READMEs, docs/api, the Lingui migration guide, troubleshooting, site steps and framework page) and tell one coherent story: application code uses the adapter, `requestKeyProvider` is adapter-only. The package wiring details are right too: `./server` subpath export plus unbuild entry, and the examples move `@palamedes/next-plugin` from devDependencies to dependencies now that it is a runtime import.
- The branch sits on the current main tip, doesn't interact with the recent lint-adapter work, and builds on the hook-free runtime model from the recent refactors instead of undercutting it. The PR description itself is a model: intent, approach with reasoning, validation list, and an honest no-follow-up rationale.

I'd approve; since this is my own PR the verdict is recorded here as a comment review instead.

packages/runtime/src/server.ts
```diff
@@ -17,35 +28,76 @@ function getServerScopeState(): ServerScopeState {
...
     },
     get() {
+      const runI18n = sharedState.getRun()
+      if (runI18n) return runI18n as T
+      const requestKey = getRequestKey?.()
+      if (requestKey) {
+        const requestScopedI18n = sharedState.requestI18n.get(requestKey)
+        if (requestScopedI18n) return requestScopedI18n as T
+      }
       return storage.getStore()
```
swernerx, line 119: The request-key lookup deliberately winning over the scope's own storage is subtle enough to deserve a one-line why-comment: a resumed or pooled context can still carry a stale instance from an earlier request's `enterWith()`, while the request key is always precise for the current request. Worth writing down so the ordering reads as a decision, not an accident. Related, one more line worth recording somewhere: with a request key present, sibling-branch `activate()` within a single request is last-write-wins — the sibling isolation test only covers the keyless scope. That's the right semantics for Next (one request, one instance, and `run()` still isolates), but it should be stated.

Issue #531, opened 2026-08-04
Automatically initialize request-scoped i18n for server functions


swernerx:
## Motivation

Palamedes can transform localized output inside server-side functions, but the translated call still needs an active request-local i18n instance. In frameworks where a server function/action is invoked as its own HTTP request, initializing i18n only at the page or component render boundary is not sufficient.

Today, applications have to remember an initialization call in every action:

```ts
export async function updateThing() {
  "use server";
  await initServerActionI18n();

  return { message: t`Saved` };
}
```

That is repetitive and easy to miss. It is especially fragile when the action itself contains no macro but calls a synchronous or asynchronous helper that does.

## Goal

Allow a Palamedes integration to initialize and activate request-scoped i18n once for every server function/action, without requiring application code to call an initializer manually in each function body.

This should be designed as a framework-neutral capability with adapter-specific implementations rather than as a Next.js-only special case.

### Shared contract

Define a small runtime/adapter contract that:

1. resolves the locale from the current request,
2. loads or obtains the catalog,
3. activates a fresh request-local i18n instance,
4. runs the complete server-function invocation inside that scope.

Applications should still own locale negotiation and catalog loading. Palamedes should own the reliable placement of that initializer/wrapper at the framework boundary.

An adapter can either wrap the request at runtime or, where the framework exposes no suitable hook, inject a configured initializer at compile time.

### Next.js: decided initial implementation

Next.js Server Actions are separate POST requests, and the public Next.js configuration currently exposes action origin/body-size settings but no general action middleware/wrapper.

Use the existing Palamedes OXC pipeline for a dedicated Server Function instrumentation pass:

```ts
withPalamedes(nextConfig, {
  serverFunctions: {
    initializer: "@/i18n/server-action#initServerActionI18n",
  },
});
```

The option is explicit and opt-in. Once configured, the transform should:

- detect inline async functions containing `"use server"`,
- detect exported async functions in modules with a top-level `"use server"` directive,
- add the initializer import once per module,
- inject `await initServerActionI18n()` after the function directive prologue and before the first executable body statement,
- instrument **every recognized Server Function**, not only functions containing a directly detected macro.

This is a separate semantic pass in the same compiler pipeline as the macro transform. Macro-driven action detection would be unsound because the translated call may live in a transitive helper or a different module.

The initializer should be request-memoized/idempotent. Synchronous and asynchronous helpers then inherit the active request scope normally:

```ts
function successMessage() {
  return t`Saved`;
}

[...]

PR #543, opened 2026-08-05
feat(next): initialize i18n in server functions

swernerx:
## Summary

- add opt-in OXC instrumentation for inline and module-level Next/React Server Functions
- initialize every recognized action before its body, including actions that only reach macros through synchronous or asynchronous helpers
- reject eager macro defaults in formal parameters and document the safe body rewrite
- exercise concurrent English/German Server Action POSTs under Turbopack and webpack

## Configuration

```js
withPalamedes({}, {
  serverFunctions: true,
})
```

The application exposes its locale and catalog initialization through a conventional
`palamedes.server` entry module. Palamedes keeps the generated import and bundler alias internal.

## Validation

- `env RUSTUP_TOOLCHAIN=stable pnpm agent:check`
- Next.js production build and concurrent RSC/Server Action scope proof with Turbopack
- the same production proof with webpack

Implements the decision-complete Next.js phase of #531. Waku and TanStack Start remain follow-up work.

Refs #531

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1172,6 +1172,7 @@ dependencies = [
  "oxc_ast_visit",
  "oxc_parser",
  "oxc_span",
+ "oxc_syntax",
  "rayon",
  "regex",
  "serde",
--- a/crates/palamedes-node/src/transform.rs
+++ b/crates/palamedes-node/src/transform.rs
@@ -4,6 +4,12 @@ use napi_derive::napi;
 use crate::shared::{checked_u32, to_napi_error};
 
 #[napi(object)]
+pub struct NativeServerFunctionTransformOptions {
+    pub initializer_module: String,
+    pub initializer_export: String,
+}
+
+#[napi(object)]
 pub struct NativeTransformOptions {
     pub runtime_module: Option<String>,
     pub runtime_import_name: Option<String>,
@@ -12,6 +18,7 @@ pub struct NativeTransformOptions {
     pub keep_source_fallbacks: Option<bool>,
     /// @deprecated Use `keepSourceFallbacks` with the inverse value.
     pub strip_message_field: Option<bool>,
+    pub server_functions: Option<NativeServerFunctionTransformOptions>,
 }
 
 #[napi(object)]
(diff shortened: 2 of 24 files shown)


swernerx:
Reviewed against #531. The core of this looks right: instrumentation is directive-driven rather than macro-driven, the initializer lands after the directive prologue, and the eager-parameter-macro hard error matches the agreed diagnostic, including the `=== undefined` guidance and the `??=` warning. The Rust test coverage is strong — concise arrow wrapping, default-export identifiers, alias collision avoidance, deferred macros in callback defaults, and the composed macro+instrumentation case are all exercised — and the concurrent EN/DE action POSTs under both bundlers give real end-to-end confidence. I also cross-checked the eager-macro lists against the visitor: excluding `Trans` (compiles to a component, resolves at render time) while treating `Plural`/`Select`/`SelectOrdinal` as eager (they compile to direct `getI18n()` calls) is correct, and `used_identifier_names` covers all scopes, so the alias collision avoidance holds beyond the top level.

One design note: the conventional `palamedes.server` entry with an internal alias is a nicer API than the `initializer: "module#export"` string that #531 records as decided. I agree with the change — worth a short note on the issue so the decision record stays accurate.

Two things worth a look before this settles:

1. Recognition gap for indirect exports in module-level `"use server"` files (`module_server_function_spans` in `crates/palamedes/src/transform/server_functions.rs`). These shapes are valid server actions but are not instrumented:

   - `export const action = withAuth(async () => { ... })` — wrapper/HOF patterns (auth wrappers, safe-action-style builders) are common in exactly these files
   - `export { action } from "./impl"` and `export * from "./impl"` when `./impl` has no directive of its own

   The failure is at least loud — the first macro hits `getI18n()` without an active scope and the action 500s with the "No active server i18n instance" error — but it only shows at request time, and nothing at build time hints at why. Since the initializer is idempotent, over-instrumenting is safe: recording any async function expression inside an exported declarator's initializer (including call arguments) would cover the wrapper case cheaply. Re-exports can't be covered from this module; a build-time warning or a documented list of recognized shapes would close that. Related: the README's "instruments every async function recognized by Next or React" slightly overpromises versus the actual recognition rules.

2. Unused macro imports are no longer stripped (`transform/mod.rs`): the import-range removal is now guarded by `!replacements.is_empty()`. A module that imports `t` but never uses it previously had the import removed; now it ships. Harmless at runtime (the macro entry only throws when called), but it's a behavior change that this PR doesn't strictly need — intentional, or a side effect of the reordering?

Neither blocks the opt-in phase in my view; item 1 is the one I'd either fix or document before promoting the flag beyond the example.

PR #545, opened 2026-08-05
feat(next): split Server Function catalogs by module

swernerx:
Depends on #543.

## What changed

- compile selected PO fragments per message-bearing server source module
- emit one lazy dynamic import per configured locale and load only the active locale
- let static ESM evaluation register fragments from transitive helpers
- keep generated catalog imports out of the browser compiler under both Turbopack and webpack
- initialize an empty request-local i18n instance first, then load the registered graph fragments
- cache locale imports across concurrent and later requests while preserving retry and HMR behavior
- document the application-facing `serverFunctions: true` setup; internal initializer module names remain hidden

## Why

The initial Server Function support in #543 loads the complete active-locale catalog inside the application initializer. This makes request initialization scale with the whole catalog even when an action and its dependency graph use only a small subset.

This stack moves catalog selection into the compile phase. In the example, the action translation lives in a transitive helper, proving that the sidecar follows normal ESM module evaluation rather than requiring the macro to live in the action file itself.

## Impact

The request still creates an isolated i18n instance, but imports only selected compiled ids for the active locale. On the example catalog this changes the action fragment from 29 messages to 1. An illustrative local Next 16.2.12 production/Webpack run measured warm English action requests at 3.10 ms median / 3.66 ms p95 versus 3.28 ms / 3.91 ms with full-catalog initialization. At this size the timing delta is small and close to noise; the main benefit is the scaling behavior.

Registration follows module evaluation, not a per-action bundler manifest. A long-lived server runtime can therefore retain registrations from multiple previously evaluated action graphs, so a later action may load a superset of its own closure. This affects the upper performance bound, not correctness: each registered module still imports only the active locale and only its selected ids.

Selected Server Function splitting currently supports PO catalogs. FCL configurations fail with an explicit build error instead of generating an unhandled import.

## Validation

- `RUSTUP_TOOLCHAIN=stable pnpm --config.verify-deps-before-run=false agent:check`
- Next 16.2.12 production build with Turbopack
- Next 16.2.12 production build with webpack
- production Server Action probes for English and German
- transitive helper sidecar verified in the generated Turbopack chunks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -65,6 +65,10 @@
 
 ### Performance Improvements
 
+- Next.js Server Functions now compile per-source message fragments and follow
+  the server ESM graph through locale-specific lazy imports. Action requests
+  load only the active locale's fragments for evaluated action modules and
+  transitive helpers instead of copying the complete locale catalog.
 - React and Solid now share one internal message renderer between their root
   and `compiled` entrypoints. This removes parallel runtime implementations
   while preserving the parser-free production boundary; lazy raw patterns use
--- a/examples/nextjs-cookie/src/lib/action-messages.ts
+++ b/examples/nextjs-cookie/src/lib/action-messages.ts
@@ -1,0 +1,7 @@
+import { t } from "@palamedes/core/macro"
+
+import type { Locale } from "./i18n"
+
+export function translateServerActionProof(locale: Locale): string {
+  return t`Server action confirmed locale ${locale}.`
+}
--- a/examples/nextjs-cookie/src/lib/actions.ts
+++ b/examples/nextjs-cookie/src/lib/actions.ts
@@ -3,7 +3,7 @@
 import { cookies } from "next/headers"
 import { revalidatePath } from "next/cache"
 import { redirect } from "next/navigation"
-import { t } from "@palamedes/core/macro"
+import { translateServerActionProof } from "./action-messages"
 import { getLocaleLabel, type Locale, LOCALES, LOCALE_COOKIE } from "./i18n"
 import { getLocale } from "./i18n.server"
 
@@ -42,8 +42,4 @@ async function createServerActionProof() {
     handledAt: new Date().toISOString(),
     message: translateServerActionProof(locale),
   }
-}
-
-function translateServerActionProof(locale: Locale) {
-  return t`Server action confirmed locale ${locale}.`
 }
--- a/examples/nextjs-cookie/src/palamedes.server.ts
+++ b/examples/nextjs-cookie/src/palamedes.server.ts
@@ -1,8 +1,12 @@
 import "server-only"
 
-import { createActiveServerI18n } from "./lib/i18n.server"
+import { createExampleI18n } from "./lib/i18n"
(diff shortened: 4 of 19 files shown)


swernerx:
The design here is convincing: fragment membership follows real ESM evaluation instead of a bundler manifest, each registered module contributes one lazy import per locale, and only the active locale materializes. The runtime tests cover exactly the behaviors that tend to bite later — concurrent dedupe, retry after a failed import, HMR replacement, the non-loadable-instance guard — and the browser/client-graph exclusion is tested for both bundlers. The transitive-helper proof in the example is the right demonstration, and the superset caveat is documented honestly rather than hidden.

Findings, roughly by weight:

1. Path canonicalization is inconsistent in `catalogMatchesSource` (`palamedes-loader.cjs`): the source path goes through `realpathSync`, but the include/exclude patterns are built from raw `config.rootDir`. On a symlinked project root (macOS `/tmp`, some CI and pnpm layouts) the patterns never match the canonicalized source, so the module silently loses splitting — a build warning appears, but at runtime the action just renders missing translations. `messageLoaderRegistration` already canonicalizes `rootDir` for the module key; doing the same here keeps both sides in agreement.

2. The passthrough guard `result.compiledIds?.length === 0` doesn't do what the `?.` suggests: if `compiledIds` were ever undefined, the comparison is false and the splitting branch reaches `JSON.stringify(undefined)`, and `Buffer.from(undefined)` throws. Either trust the transform contract and drop the `?.`, or invert to `!result.compiledIds?.length`.

3. Modules first evaluated after the initializer already ran — a dynamic `import()` inside the action body — register their fragments too late for the current request: that request renders those messages untranslated, the next one works. That's inherent to the static-graph design, but the per-request flakiness is easy to misdiagnose. One explicit sentence in the README ("dynamic imports inside a handler register for subsequent requests only") would save someone a debugging session.

4. README snippet drift: the `palamedes.server` example imports `serverI18n` and does `const locale = await getLocale()`, but the actual example app uses `serverI18nScope` and destructures `const { locale } = await getLocale()`. Since this snippet is the copy-paste setup path, aligning it with `examples/nextjs-cookie` avoids immediate friction.

5. `catalogMatchesSource` re-implements catalog include/exclude semantics (directory-to-`**/*.{js,jsx,ts,tsx,mdx}` expansion, default excludes) inside a shipped `.cjs` file. If extraction-side matching ever changes, this copy drifts silently. If the matcher can be exposed from `@palamedes/config` or core-node, reusing it is worth it; otherwise cross-linking the two sites in comments would do. Optional.

One observation, no action needed: SSR copies of client components also carry registrations in the server graph, so a warmed runtime accrues page-component fragments that action requests then load — the documented superset effect in practice. If the upper bound ever matters, that's where it will come from first.

Also checked, all fine: pseudoLocale is validated to be part of `locales`, so the per-locale loader map covers it; the generated `.then(({ messages }) => messages)` matches the `export const messages=` shape of rendered catalog modules; and `failOnMissing` flows through `createCatalogLoaderResult` in the selected path the same way as in the full-catalog path.

Still marked draft — from my side, items 1 and 4 are the ones to pick up before undrafting; the rest can trail.

PR #546, opened 2026-08-05
feat(next): split client messages by module graph

### LINKED ISSUE #544: feat(next): replace app-owned catalog boundaries with graph-split bootstrap
## Problem

Palamedes 1.12 introduced the render-safe `createClientCatalogBoundary()` in #509 (from #507). It solves the immediate Next.js App Router bootstrap problem correctly:

- only the active locale's full catalog is loaded on the client;
- the parser-free client instance is initialized before translated descendants hydrate;
- the locale is fixed for the lifetime of the browser document;
- a locale change requires a document navigation.

The remaining application API is deliberately low-level. Every Next app must still create a `"use client"` module that contains:

1. a bundler-visible dynamic `.po` import;
2. a document-locale resolver and runtime validation;
3. an exported boundary factory result;
4. a root-layout wrapper.

A real migration now contains near-identical versions of that module in both its Web and Admin apps: [3DegreesGroupInc/EACManagement#1673](https://github.com/3DegreesGroupInc/EACManagement/pull/1673).

We could shorten the factory API and default the locale resolver to `document.documentElement.lang`. That would improve the current DX, but it risks polishing an abstraction that should disappear once Palamedes' graph-splitting work reaches Next.js.

The current boundary is Stage 1 / V1 from the [code-splitting RFC](https://github.com/sebastian-software/palamedes/blob/main/docs/plans/2026-08-01-code-splitting-localization-rfc.md): one lazy chunk containing the active locale's complete catalog. The experimental Vite implementation has already shown the next step: messages follow the code/module graph and register before the consuming code renders. For Next/RSC, the RFC identifies a server-first delivery shape as the likely end state.

## Goal

Bring automatic message graph splitting to `@palamedes/next-plugin` and replace the application-owned full-catalog boundary with a Palamedes-owned, generic client bootstrap.

The target developer experience should not require an application to:

- create a custom catalog boundary;
- import a complete client catalog;
- repeat `<html lang>` resolution or validation;
- manually partition messages by route or namespace;
- serialize executable catalog functions through RSC props.

The application should continue to own locale negotiation and set the document locale. Palamedes should own build-time message attribution, client delivery, and initialization ordering.

### Build time

Reuse the model proven by `experimentalGraphSplitting` in the Vite plugin:

- retain `compiledIds` for transformed Client Component modules;
- compile only the catalog entries used by the relevant client graph;
- preserve the branded, executable, parser-free catalog ABI;
- derive membership from the bundler/framework graph rather than author-defined namespaces;
- emit deterministic artifacts and retain the existing missing-translation and fallback behavior.

The Next implementation does not have to copy the Vite delivery mechanism. In particular, import-map injection may conflict with Next/React module-preload ordering. The implementation should first spike the supported Next and Turbopack extension points and choose a delivery shape that is reliable for App Router production builds.

Likely options include:

- generated sidecar registration modules that follow Client Component chunks;
- per-segment/per-locale message assets selected by the server;
- a server-computed initial message subset plus message deltas delivered with client navigations.

Any RSC-oriented design must explicitly solve the compiled-ABI constraint: generated messages are executable functions and must not be naively serialized as ordinary RSC/JSON props.

### Runtime and hydration

For the initial document:

1. the server resolves and activates its request-local i18n instance as today;
2. only messages required by the initial Client Component graph are delivered for the active locale;
3. Palamedes creates and activates the client instance before translated descendants hydrate;
4. generated registrations that evaluated before installation are flushed into that instance.

For a client-side route/segment navigation, newly required messages must be available before the corresponding translated components render.

The bootstrap should be generic and owned by Palamedes. If a React boundary remains as an internal implementation detail, applications should consume one ready-made component rather than construct a catalog loader factory.

Locale lifecycle remains unchanged: one locale per browser document, with document navigation for changes. This issue must not reintroduce reactive/live locale switching.

## Fallback and migration

The current `createClientCatalogBoundary()` remains a valid compatibility path for:

- Next versions or bundlers without the necessary graph hooks;
- applications that deliberately choose one complete active-locale catalog;
- other hosts where automatic graph splitting is unavailable.

Once the Next integration is proven, documentation and examples should present graph splitting as the preferred production path and the full-catalog boundary as the explicit fallback. Avoid investing in a separate application wrapper or code-generated source file solely to hide the transitional factory API.

## Non-goals

- Live locale switching inside a mounted document.
- Author-facing namespaces or route-specific catalog files.
- Changing the translator workflow or splitting checked-in `.po` catalogs.
- Shipping all locales in every route chunk as the final production design.
- Passing executable catalog functions through Client Component props.
- Solving request-scoped Server Function initialization (#531) or the independent server-scope suspension issue (#538).

## Acceptance criteria

- [ ] A production Next.js App Router example uses automatic message splitting without an application-created `createClientCatalogBoundary()` module.
- [ ] The example imports no complete `.po` catalog from its client graph.
- [ ] The initial browser payload contains only the active locale's messages needed by the initial Client Component/segment graph.
- [ ] Messages belonging only to another locale or unrelated lazy route are absent from the initial payload.
- [ ] Client-side navigation delivers newly required messages before their translated components render.
- [ ] Initial SSR and hydration render translated Client Components correctly without missing-runtime or missing-message fallbacks.
- [ ] The client bootstrap does not depend on effect timing or speculative render-phase mutation.
- [ ] Locale/document mismatches fail clearly, and locale changes remain full document navigations.
- [ ] Generated client messages remain branded, precompiled, and parser-free.
- [ ] Fallback locale resolution, pseudo-locale behavior, `failOnMissing`, and `failOnCompileError` remain correct for split artifacts.
- [ ] Catalog edits invalidate the relevant development artifacts and update through HMR or a documented reload path.
- [ ] Next 16 + React 19 + Turbopack production behavior is covered by an integration test.
- [ ] Webpack support is either covered by the same contract or documented as using the full-catalog fallback.
- [ ] Bundle/network assertions prove the active-locale × current-client-graph payload property.
- [ ] The Next examples, React/Next API docs, and the code-splitting RFC are updated to describe the final supported contract.

## Investigation questions

- Which stable Next/Turbopack hooks expose enough transform and module-graph information to generate or associate message sidecars?
- Can message registrations ride directly with Next client chunks, or must Next emit and select separate per-locale assets?
- How can the active locale bind before React/Next module preloads execute without unsafe HTML ordering?
- How are message deltas associated with App Router client navigations and RSC segment payloads?
- Should the generic bootstrap be public API, generated adapter code, or entirely internal to the Next integration?
- What is the smallest reliable fallback when a Next/Turbopack version does not expose the required graph lifecycle?

## Related

- #507 — original render-safe client catalog bootstrap problem
- #509 — current full-catalog boundary implementation
- #531 — separate Server Function request initialization
- #538 — separate Next RSC server-scope lifetime problem
- [Code-splitting localization RFC](https://github.com/sebastian-software/palamedes/blob/main/docs/plans/2026-08-01-code-splitting-localization-rfc.md)
- [3Degrees EAC migration PR #1673](https://github.com/3DegreesGroupInc/EACManagement/pull/1673)

swernerx:
## Summary

- add the public messageSplitting flag and generate browser-only, active-locale PO subset imports per translated client module
- initialize one shared parser-free client runtime before module evaluation, while leaving Server Components and Server Functions on their existing request-local paths
- remove the application-owned catalog boundary from the Next reference example and prove lazy client navigation across Turbopack and webpack
- document PO support, parser-free authoring constraints, and the complete-catalog compatibility fallback

## Verification

- Turbopack Next 16 production build, RSC and Server Action concurrency probes, Chromium hydration, and locale x route network assertions
- webpack Next 16 production build with the same browser and network assertions
- env CI=true RUSTUP_TOOLCHAIN=stable pnpm --config.verify-deps-before-run=false agent:check

Closes #544


packages/next-plugin/palamedes-loader.cjs
```diff
@@ -77,14 +77,61 @@ function catalogResourcePath(config, catalog, locale) {
...
+    return `{ ${loaders} }`
+  })
+  const supportedLocales = config.locales.map((locale) => JSON.stringify(locale)).join(", ")
+  const modulePath = normalizePath(
+    path.relative(canonicalPath(config.rootDir), canonicalPath(sourcePath))
+  )
+  const identifier = `__pmds_${createHash("sha256").update(modulePath).digest("hex").slice(0, 12)}`
+
+  return (
+    `\nconst ${identifier}_locale = document.documentElement.lang;\n` +
```
swernerx, line 130: One boundary worth pinning down: because the bootstrap is appended after the module body, anything evaluated at module scope — `const label = t(...)` outside a component — still runs before the fragment (or even the client instance) exists, and dies with the generic missing-instance error. Appending is the right call for keeping the source map valid, so if module-scope translation calls are already outside the supported contract, all good — but if that isn't written down yet, one sentence in the next-plugin docs would save the first person who hits this a confusing session. Not blocking.

packages/next-plugin/palamedes-loader.cjs
```diff
@@ -77,14 +77,61 @@ function catalogResourcePath(config, catalog, locale) {
...
+function clientMessageBootstrap(config, sourcePath, compiledIds) {
+  const catalogs = config.catalogs.filter((catalog) =>
+    catalogMatchesSource(config, catalog, sourcePath)
+  )
+  if (catalogs.length === 0) {
+    return null
+  }
+
+  const selection = Buffer.from(JSON.stringify(compiledIds)).toString("base64url")
+  const loaderGroups = catalogs.map((catalog) => {
```
swernerx: Small and optional: this is the second copy of the catalog-filter → selection → per-locale-specifier shape in this file (`messageLoaderRegistration` builds identical specifiers, just with a `.then` unwrap). A shared helper returning the per-catalog locale→specifier map would keep the two generators from drifting — the query name and encoding have to stay in sync for the po-loader to pick them up. Take it or leave it.

packages/next-plugin/src/index.test.ts
```diff
@@ -145,6 +145,27 @@ describe("withPalamedes turbopack config", () => {
...
+      conditionList(candidate).some(
+        (condition) =>
+          typeof condition === "object" &&
+          condition !== null &&
+          "not" in condition &&
+          condition.not === "browser"
+      )
+    )
+
+    expect(browserRule?.loaders?.[0]?.options).toMatchObject({ clientMessageSplitting: true })
```
swernerx, line 165: Worth one more assertion here: this PR is what made `serverMessageSplitting` conditional in the non-browser rule (it was hardcoded before), and with only `messageSplitting` enabled nothing currently locks in that the server rule stays free of it. A regression would silently re-enable server splitting without the initializer wiring. One extra `.not.toHaveProperty("serverMessageSplitting")` on `serverRule` would pin the contract.

swernerx (commented):
Reviewed the full change against #544. A few things here are genuinely well done and worth naming:

- The delivery shape — a per-module top-level-await bootstrap over statically enumerable per-locale subset imports — reuses the proven server-splitting query mechanism instead of inventing an import-map analog, and it keeps every hard constraint from the ticket: no executable messages through RSC, no inline script, locale fixed per document, parser-free ABI intact.
- `initializeClientI18n` reusing a matching already-installed instance means graph-split modules and the compat boundary can coexist in one document, and going through `setClientI18n` on first install covers the buffered-registration flush from the ticket for free.
- The e2e proof is strong: chunk-body assertions for active and inactive sentinels across both bundlers, and a hidden `router.push` probe instead of a `<Link>` so prefetch can't pull the lazy fragment into the initial phase and fake a pass. The `de` navigation assertion also catches a source-fallback render, which the `en` case alone couldn't.

Nothing blocking. Two open questions since this closes #544:

1. Module-scope translation calls — see the inline comment on the loader. Supported, or documented-out?
2. The ticket's dev/HMR criterion (catalog edits invalidate development artifacts and update through HMR or a documented reload path): the subset imports should invalidate through normal dependency tracking, but I don't see it verified or documented anywhere in the PR. Was that checked manually, or should it get a doc line / follow-up?

One small doc note without a diff anchor: `examples/README.md` ("Catalog Loading And Bundle Size") still presents dynamic imports plus `createClientCatalogBoundary()` as the Next answer. Now that `nextjs-cookie` demonstrates `messageSplitting`, a sentence pointing there as the preferred path would match the API docs and the RFC update. Fine as a follow-up.

GitHub won't let the author account file a formal approval, so recording it here instead: merge-ready from this review once the two questions are settled.

PR #594, opened 2026-08-07
fix(core-node): catch panics at N-API boundary

### LINKED ISSUE #570: fix(core-node): panics cross the N-API boundary as a process abort — catch_unwind is never enabled
## Problem

None of the sync exports in `crates/palamedes-node` uses `#[napi(catch_unwind)]` (repo grep: zero hits). With the pinned napi 3.12.0 / napi-derive-backend 6.1.1, the generated wrapper is a plain `extern "C" fn` and `std::panic::catch_unwind` is emitted **only** when the export opts in (`napi-derive-backend-6.1.1/src/codegen/fn.rs:154`, `:320`).

`packages/core-node/scripts/build-native.mjs:120-123` documents the opposite belief: "napi-rs (`#[napi]`) wraps every `extern \"C\"` entry point in a panic guard, so panics are converted to JS errors and never unwind across the FFI boundary" — true for async paths and v2-era behavior, false for these v3 sync exports.

Reachable panic sites include `crates/palamedes/src/translation_candidates.rs:850` (`.expect`), `:1017` (`unreachable!`), `crates/palamedes-node/src/catalog.rs:2072` (`.expect`), plus anything inside oxc/ferrocat.

## Impact

Any panic in the ~23 sync exports unwinds into an `extern "C"` frame → Rust aborts the process (SIGABRT). A Vite/Next dev server or CI build dies without a catchable JS error — exactly the failure mode the build-script comment claims cannot happen.

## Direction

Add `catch_unwind` to every export (or a crate-local macro wrapper), and correct the build-native.mjs comment so the false invariant does not justify future decisions (it is currently cited as the reason `panic=abort` is not forced).

swernerx:
## Summary

- Opt every synchronous `palamedes-node` N-API export into `catch_unwind`.
- Add a regression test that rejects unguarded synchronous exports.
- Correct the native-build panic-safety comment.

## Issue

Closes #570

## Validation

- `rustup run 1.95-aarch64-apple-darwin cargo test -p palamedes-node --locked`
- `rustup run 1.95-aarch64-apple-darwin cargo clippy -p palamedes-node --all-targets --locked -- -D warnings`
- `pnpm format:check`
- `RUSTUP_TOOLCHAIN=1.95-aarch64-apple-darwin pnpm --filter @palamedes/core-node-darwin-arm64 build`
- `RUSTUP_TOOLCHAIN=1.95-aarch64-apple-darwin pnpm --filter @palamedes/core-node check-native-types`
- `pnpm --filter @palamedes/core-node test`

## Follow-up

None.

### DIFF
--- a/crates/palamedes-node/src/lib.rs
+++ b/crates/palamedes-node/src/lib.rs
@@ -30,7 +30,7 @@ mod tests {
             let unguarded_exports = napi_attributes(&source)
                 .into_iter()
                 .filter(|(_, attribute)| {
-                    !is_object_or_enum_annotation(attribute) && !attribute.contains("catch_unwind")
+                    !is_object_or_enum_annotation(attribute) && !has_catch_unwind_option(attribute)
                 })
                 .map(|(line, _)| line)
                 .collect::<Vec<_>>();
@@ -56,6 +56,7 @@ mod tests {
             vec![(1, "#[napi(js_name = \"exported\")]".to_owned())]
         );
         assert!(!is_object_or_enum_annotation(&attributes[0].1));
+        assert!(!has_catch_unwind_option(&attributes[0].1));
     }
 
     #[test]
@@ -69,6 +70,31 @@ mod tests {
         assert!(attributes
             .iter()
             .all(|(_, attribute)| is_object_or_enum_annotation(attribute)));
+    }
+
+    #[test]
+    fn napi_option_scan_ignores_string_and_comment_collisions() {
+        let unguarded_attributes = [
+            format!("#[{}]", r#"napi(js_name = "catch_unwind")"#),
+            format!("#[{}]", r#"napi(ts_return_type = "catch_unwind")"#),
+            format!("#[{}]", r##"napi(ts_return_type = r#"catch_unwind"#)"##),
+            format!("#[{}]", "napi(\njs_name = \"exported\", // catch_unwind\n)"),
+            format!("#[{}]", "napi(/* catch_unwind */ js_name = \"exported\")"),
+        ];
+
+        assert!(unguarded_attributes
+            .iter()
+            .all(|attribute| !has_catch_unwind_option(attribute)));
+    }
+
+    #[test]
+    fn napi_option_scan_recognizes_multiline_catch_unwind_option() {
+        let attribute = format!(
+            "#[{}]",
+            "napi(\njs_name = \"exported\",\n// Keep panics inside the N-API boundary.\ncatch_unwind,\nts_return_type = \"string\"\n)"
+        );
+
+        assert!(has_catch_unwind_option(&attribute));
     }
 
     fn rust_source_modules(dir: &Path) -> Vec<PathBuf> {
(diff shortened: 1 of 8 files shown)


crates/palamedes-node/src/lib.rs
```diff
@@ -13,3 +13,31 @@ pub use self::mdx::*;
...
+        ("transform.rs", include_str!("transform.rs")),
+    ];
+
+    #[test]
+    fn every_sync_napi_export_enables_panic_catching() {
+        for (module, source) in SYNC_NAPI_EXPORT_MODULES {
+            let unguarded_exports = source
+                .lines()
+                .enumerate()
+                .filter_map(|(index, line)| (line.trim() == "#[napi]").then_some(index + 1))
```
swernerx: [P2] Make this check cover every function-level N-API attribute. The hard-coded list already omits catalog_config.rs, so a future synchronous export in that existing module (or a new module) would not be scanned. Within scanned files, matching only the exact bare #[napi] line also lets an unguarded #[napi(js_name = "...")] export through. Either path can reintroduce the process-abort failure while this test stays green, which misses the regression coverage required by #570. Please cover every Rust source module and reject every function or method-level #[napi(...)] attribute that lacks catch_unwind, while excluding object and enum annotations.

swernerx (commented):
The annotations now cover all 23 current synchronous exports, and the corrected build comment matches napi-rs 3.12.0 generated guard. One blocking regression-test gap remains inline. GitHub does not allow requesting changes on a PR authored by the active account, so this is submitted as a comment. Verdict: changes requested.

crates/palamedes-node/src/lib.rs
swernerx: Addressed in f519da8. The regression check now recursively scans every Rust module under `crates/palamedes-node/src` (including `catalog_config.rs` and future modules), rejects each `#[napi]` or `#[napi(...)]` attribute without `catch_unwind`, and excludes only `#[napi(object...)]` and `#[napi(string_enum...)]` type annotations. Added focused coverage for `#[napi(js_name = "...")]` and the object/enum exclusions.

Validated with:
- `rustup run 1.95-aarch64-apple-darwin cargo test -p palamedes-node --locked` (3 passed)
- `rustup run 1.95-aarch64-apple-darwin cargo clippy -p palamedes-node --all-targets --locked -- -D warnings`
- `RUSTUP_TOOLCHAIN=1.95-aarch64-apple-darwin pnpm --filter @palamedes/core-node check-native-types`
- `pnpm --filter @palamedes/core-node test` (18 passed)

crates/palamedes-node/src/lib.rs
```diff
@@ -13,3 +13,109 @@ pub use self::mdx::*;
...
+    #[test]
+    fn every_function_napi_export_enables_panic_catching() {
+        let source_dir = Path::new(env!("CARGO_MANIFEST_DIR")).join("src");
+
+        for module in rust_source_modules(&source_dir) {
+            let source = fs::read_to_string(&module).expect("read Rust source module");
+            let unguarded_exports = napi_attributes(&source)
+                .into_iter()
+                .filter(|(_, attribute)| {
+                    !is_object_or_enum_annotation(attribute) && !attribute.contains("catch_unwind")
```
swernerx: [P2] Detect catch_unwind as an attribute option, not as arbitrary text. For example, #[napi(js_name = "catch_unwind")] is a valid synchronous export with no panic guard, but this contains check treats it as guarded and the regression test stays green. String-valued options such as ts_return_type, or even an attribute comment, create the same bypass, so the process-abort failure can still be reintroduced undetected. Please parse or tokenize the top-level napi options and add the js_name collision case to the focused tests.

swernerx (commented):
The recursive module scan now covers all Rust sources, and the object/string_enum exclusions match the current type annotations. One blocking guard-detection bypass remains inline. GitHub does not allow requesting changes on my own PR, so this is submitted as a comment. Verdict: changes requested.

crates/palamedes-node/src/lib.rs
swernerx: Addressed in 8546a2d. The guard check now tokenizes top-level `#[napi(...)]` options, skipping ordinary/raw strings, line comments, block comments, and nested delimiters. `catch_unwind` is recognised only when it is an option identifier; object/string-enum detection uses the same tokenization.

Added adversarial coverage for `js_name = "catch_unwind"`, ordinary and raw `ts_return_type` strings, line/block-comment collisions, plus a commented multiline real `catch_unwind` option.

Validated with:
- `rustup run 1.95-aarch64-apple-darwin cargo test -p palamedes-node --locked` (5 passed)
- `rustup run 1.95-aarch64-apple-darwin cargo clippy -p palamedes-node --all-targets --locked -- -D warnings`
- `RUSTUP_TOOLCHAIN=1.95-aarch64-apple-darwin pnpm --filter @palamedes/core-node check-native-types`
- `pnpm --filter @palamedes/core-node test` (18 passed)

swernerx (commented):
Re-reviewed at 8546a2d. Both prior blockers are resolved: the check recursively covers every Rust source module, and the tokenizer only recognizes catch_unwind as a top-level N-API option while ignoring string, comment, and nested-delimiter collisions. The current object/string_enum exclusions and all 23 guarded synchronous exports remain correct. GitHub does not allow approving my own PR, so this is a comment. Verdict: clear.

PR #597, opened 2026-08-07
fix(react): evict rejected catalog resources

### LINKED ISSUE #571: fix(react): client catalog boundary caches catalog-load rejections forever (client and server)
## Problem

In `packages/react/src/client.tsx`:

- `:59-64` — `clientI18nResource` is created at module evaluation with no `.catch` and is never replaced; every render `use()`s the same promise (`:83-86`), so a rejection re-throws forever.
- `:66-73` — `serverResources` caches a rejected promise per locale for the process lifetime.

The runtime's own lazy loaders already do this correctly — they evict on failure (`packages/runtime/src/index.ts:177-181`).

## Impact

One transient network failure fetching the catalog chunk permanently breaks translated rendering for that document (only a full reload recovers), and before the boundary first renders the rejection surfaces as an unhandled promise rejection. On the server, one failed `loadCatalog` poisons that locale for the lifetime of a long-running process.

## Direction

On rejection, evict/replace the cached promise (client: retry on next render; server: `serverResources.delete(locale)` in a `.catch` that rethrows), mirroring `runtime/index.ts:177-181`.

swernerx:
## Summary

- Evict rejected client and server catalog resources while preserving eager client preloads and successful cache reuse.
- Observe client preload rejections until React consumes them, avoiding unhandled rejection behavior.
- Add client and server regression coverage for recovery, repeated failures, original error propagation, and cache reuse.

## Issue

Closes #571

## Validation

- `pnpm --filter @palamedes/react... build`
- `pnpm --filter @palamedes/react test`
- `pnpm --filter @palamedes/react exec tsc --noEmit -p tsconfig.json`
- `pnpm format:check`
- `pnpm lint`
- `RUSTUP_TOOLCHAIN=1.95-aarch64-apple-darwin pnpm test`
- `RUSTUP_TOOLCHAIN=1.95-aarch64-apple-darwin pnpm check-types`
- `pnpm proof:icu:check`
- `pnpm check:release-set`

## Follow-up

None.

### DIFF
--- a/packages/react/src/client.tsx
+++ b/packages/react/src/client.tsx
@@ -58,7 +58,7 @@ export function createClientCatalogBoundary<TLocale extends string>({
   const clientLocale = typeof window === "undefined" ? undefined : resolveClientLocale()
   let clientI18nResource: Promise<CompiledPalamedesI18n> | undefined
   let clientI18nResourceRejected = false
-  let clientI18nResourceEvictionScheduled = false
+  let clientI18nResourceError: unknown
 
   function createClientI18nResource(locale: TLocale): Promise<CompiledPalamedesI18n> {
     const resource = loadCatalog(locale).then((catalog) =>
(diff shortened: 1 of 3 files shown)


packages/react/src/client.tsx
```diff
@@ -56,18 +56,42 @@ export function createClientCatalogBoundary<TLocale extends string>({
...
+  let clientI18nResource: Promise<CompiledPalamedesI18n> | undefined
+
+  function createClientI18nResource(locale: TLocale): Promise<CompiledPalamedesI18n> {
+    const resource = loadCatalog(locale).then((catalog) =>
+      setClientI18n(createCatalogI18n(locale, catalog))
+    )
+    clientI18nResource = resource
+    void resource.catch(() => {
+      if (clientI18nResource === resource) {
+        clientI18nResource = undefined
```
swernerx: [P1] Preserve the rejected resource until React consumes it

This clears `clientI18nResource` in the promise rejection microtask, before React's suspended render retries. The retry calls `getClientI18nResource()` and starts a new promise instead of synchronously rethrowing the rejected one to the nearest error boundary. In a minimal React 19/jsdom repro with two rejected loads followed by a pending load, `loadCatalog` was called 3 times, the UI stayed on the Suspense fallback, and `componentDidCatch` saw 0 errors. A persistent outage can therefore keep retrying without ever reaching the app's error UI. Please keep the rejected promise available until `use(resource)` surfaces its original error, evict it as part of that consumption path, and add the repeated-failure/error-boundary regression test. The eager observer can remain separate so pre-render rejections stay handled.

swernerx (commented):
Verdict: changes requested. The eager preload timing and identity-guarded cache work are headed in the right direction, but one blocking client error-path issue remains inline. Clearing the rejected resource before React consumes it keeps the rejection away from the app's error boundary and can turn a persistent outage into repeated catalog loads under Suspense. I'm submitting COMMENT because GitHub doesn't allow the PR author account to request changes on its own PR.

packages/react/src/client.tsx
swernerx: Fixed in `9d528465405b3c6c6c42c836691bf8551f70dd61`.

The eager rejection observer now records a handled rejection while retaining the resource. When the boundary consumes that rejected resource, it schedules eviction for after React can commit the nearest error boundary, then calls `use(resource)` unchanged so the original error is surfaced. A later retry therefore starts a fresh load instead of looping in Suspense.

Added regression coverage for a rejection before the first render, two surfaced failures with load counts 1 and 2, then successful retry at count 3; the existing eager-preload and successful/inflight identity coverage remains. Verified with focused React build/tests, full workspace tests and typechecks, format, lint, published-types, ICU proof, and release-set checks.

🔧 Mira · Implementation Agent

swernerx: Superseding my earlier reply: the final fix is `c99725350fd0ca48a4d09fa589dc57191130fac4`.

CI exposed a race in the initial deferred-eviction approach, so the cache now evicts only from an internal boundary’s `componentDidCatch` after React has caught the original rejected resource. That boundary then rethrows the same error to the application’s nearest error boundary. The eager observer remains separate and marks pre-render rejections handled without eviction.

The repeated-failure regression now passes ten consecutive focused runs, plus the full workspace test/typecheck/format/lint/published-types/ICU/release-set validation. New CI is running.

🔧 Mira · Implementation Agent

swernerx (commented):
Verdict: clear

The prior P1 is fully addressed at this head. The eager observer retains the rejected promise until React catches it; the internal boundary evicts only from componentDidCatch, then rethrows the same original error to the application boundary. The focused regression and an independent React 19 Strict Mode/concurrent harness both preserve one load per retry, in-flight and successful reuse, and original error identity without unhandled rejection or automatic retry loops. Server locale eviction and successful caching remain intact. All seven CI checks are green.

PR #598, opened 2026-08-07
fix(core): fingerprint complete translation origins

### LINKED ISSUE #568: fix(core): patch fingerprints are computed over the truncated origin list — messages with more origins than maxOrigins can never be patched
## Problem

`build_candidate` truncates origins with `.take(max_origins)` **before** computing the candidate fingerprint, while the apply side rebuilds the candidate with `usize::MAX` and compares fingerprints:

- `crates/palamedes/src/translation_candidates.rs:696` (truncation), `:720`/`:832-847` (fingerprint over `candidate.origins`)
- `crates/palamedes/src/translation_candidates.rs:475-484` (apply side: `build_candidate(&loaded, &key, message, usize::MAX)` then reject on mismatch)
- Default `maxOrigins` is 8 (`crates/palamedes-node/src/catalog.rs:799`); the docs example even recommends `maxOrigins: 5` (`docs/translation-candidate-patches.md:78`)

Reproduced against the current crate: a PO entry with 10 `#:` origins, listed with the default `max_origins`, then patched with the returned fingerprint → `outcome: Rejected`, `translation.stale_candidate`, `updated: false` — with nothing changed on disk.

## Impact

Any message referenced from more files than `maxOrigins` (shared strings like "Save") is enumerated but permanently un-patchable. Providers land in a stale-candidate rejection loop that re-listing never resolves — the entire documented translate-and-write-back loop blocks. The core-node test only uses 1 origin, so this is uncovered.

## Direction

Exclude origins from the fingerprint (they are presentation data, not concurrency state), or compute the fingerprint before truncation so both sides hash the full list. Also dedupe the default `8` (binding hardcodes it; core has `DEFAULT_MAX_ORIGINS`). Note: this changes fingerprint values, so in-flight candidates must be re-listed.

*Filed from the 2026-08-07 repository audit (two independent reviews reproduced this by execution).*

swernerx:
## Summary

- hash complete catalog origin state rather than the bounded candidate response list
- share the core default origin limit with the Node binding
- cover truncated candidate patching in Rust and core-node, and document the v2 fingerprint relist requirement

## Issue

Closes #568

## Validation

- `rustup run 1.95 cargo test --workspace --locked`
- `rustup run 1.95 cargo clippy --workspace --all-targets --locked -- -D warnings`
- `pnpm build`
- `pnpm check-types`
- `pnpm format:check`
- `pnpm lint`
- `pnpm --filter @palamedes/core-node test`

## Follow-up

None.

### DIFF
--- a/crates/palamedes/src/translation_candidates.rs
+++ b/crates/palamedes/src/translation_candidates.rs
@@ -97,7 +97,7 @@ pub struct TranslationCandidate {
 }
 
 /// Source origin exposed to translation workflow consumers.
-#[derive(Clone, Debug, PartialEq, Eq, Serialize)]
+#[derive(Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct TranslationWorkflowOrigin {
     /// Source filename.
@@ -694,7 +694,7 @@ fn build_candidate(
         || message.comments.clone(),
         |item| item.extracted_comments.iter().cloned().collect(),
     );
-    let all_origins = message
+    let mut all_origins = message
         .origin
         .iter()
         .map(|origin| TranslationWorkflowOrigin {
@@ -702,6 +702,8 @@ fn build_candidate(
             scope: origin.scope.clone(),
         })
         .collect::<Vec<_>>();
+    all_origins.sort();
+    all_origins.dedup();
     let id = TranslationCandidateId {
         catalog: loaded.scope.clone(),
         locale: loaded.locale.clone(),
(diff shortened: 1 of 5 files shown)


crates/palamedes/src/translation_candidates.rs
```diff
@@ -690,15 +694,14 @@ fn build_candidate(
...
+    let all_origins = message
         .origin
         .iter()
-        .take(max_origins)
         .map(|origin| TranslationWorkflowOrigin {
             file: origin.file.clone(),
             scope: origin.scope.clone(),
         })
-        .collect();
+        .collect::<Vec<_>>();
```
swernerx, line 704: [P2] Canonicalize the full origin set before hashing

`all_origins` still preserves raw PO/FCL order and duplicates. I reproduced the same logical origin set as `b, a, b` and `a, b`; v2 returned different fingerprints, so a semantic no-op reorder or deduplication can still trigger `translation.stale_candidate`. Please sort and deduplicate by `(file, scope)` before truncating the response and hashing, and cover reordered/duplicate origins in the regression test.

swernerx (commented):
Verdict: changes requested

crates/palamedes/src/translation_candidates.rs
swernerx: Addressed in a68cd4e9e9dcf5f984ee5bf926d83e784f4dce32. Complete origins are now sorted and deduplicated by `(file, scope)` before both response truncation and fingerprinting.

Regression coverage starts with >8 reversed, duplicated origins, then rewrites the same logical set in canonical order and verifies identical returned origins/fingerprints; it also retains the hidden-origin invalidation and truncated-candidate patch checks.

Validated with `rustup run 1.95 cargo test --workspace --locked`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `pnpm --filter @palamedes/core-node test`, generated native type checks, `pnpm format:check`, and `pnpm lint`.

🔧 Mira · Implementation Agent

swernerx (commented):
Verdict: clear

PR #600, opened 2026-08-07
fix(transform): instrument local server action handlers

### LINKED ISSUE #575: fix(transform): server actions wrapped around locally referenced handlers are not instrumented
## Problem

In a module-level `"use server"` file, `module_server_function_spans` (`crates/palamedes/src/transform/server_functions.rs:91-150`) records only (a) exported function declarations/initializers and (b) local declarations whose own name is exported. `record_expression_function` (`:203-228`) records only async functions syntactically nested in an exported initializer.

For the common wrapper-by-reference pattern:

```ts
"use server"
async function handler() { return t`...` }
export const save = withAuth(handler)
```

`handler` is neither exported by name nor nested in the initializer → no `await __palamedesServerFunctionInitializer()` is injected anywhere.

`packages/next-plugin/README.md:220-222` documents only the *imported*-callback limitation, implying local bodies are covered — they are not when passed by reference.

## Impact

Wrapper patterns (`withAuth(handler)`, next-safe-action style) produce actions that run without request i18n; the first macro inside throws "No active server i18n instance" → 500 at request time, with no build-time hint.

## Direction

When an exported initializer contains a `CallExpression`, resolve identifier arguments against module-local async function declarations/consts and record their spans (over-instrumenting a local helper only costs an idempotent initializer call). Update the README sentence either way.

swernerx:
## Summary

- Resolve module-local async function declarations and immutable async const function/arrow bindings passed by identifier into exported Server Function wrappers.
- Preserve wrapper nesting, aliases, span deduplication, and shadowing/import safety.
- Document locally referenced callback support and add transform regressions.

## Issue

Closes #575

## Validation

- cargo +1.95 fmt --all --check
- cargo +1.95 test --workspace --locked
- cargo +1.95 clippy --workspace --all-targets --locked -- -D warnings
- pnpm format:check
- pnpm lint
- RUSTUP_TOOLCHAIN=1.95 pnpm build
- RUSTUP_TOOLCHAIN=1.95 pnpm test
- RUSTUP_TOOLCHAIN=1.95 pnpm check-types

## Follow-up

None.

🔧 Mira · Implementation Agent


crates/palamedes/src/transform/server_functions.rs
```diff
@@ -215,15 +333,39 @@ impl<'a> Visit<'a> for ExportedInitializerFunctionCollector<'_> {
...
+            for argument in &it.arguments {
+                let Some(Expression::Identifier(identifier)) = argument
+                    .as_expression()
+                    .map(Expression::get_inner_expression)
+                else {
+                    continue;
+                };
+                if let Some(span) = self
+                    .local_async_function_spans
+                    .get(identifier.name.as_str())
```
swernerx: This lookup matches by spelling, not binding identity. For example, `async function handler() {}; export const save = wrap(class Config { static { const handler = () => {}; register(handler); } })` instruments the outer `handler` even though the call refers to the class-local binding. Function declarations are also reassignable (`handler = async () => ...`), so the table can instrument the old body while the wrapper receives an uninstrumented replacement. Both cases can initialize request i18n in an unrelated helper and still leave the actual action uncovered. Please resolve identifier references to their module binding and exclude bindings with writes (or otherwise track the effective target), with regressions for both cases.

swernerx (commented):
Verdict: changes requested

The direct module-local declaration and immutable const handler cases are covered cleanly, including nested wrappers, export aliases, span deduplication, and the existing exported-initializer behavior. The focused transform suite, formatting, clippy, and an independent syntax/source-map/determinism harness all pass for those paths.

One binding-safety issue blocks this head: reference resolution is name-based, so it does not distinguish non-function nested bindings or writes to function-declaration bindings. The inline comment has concrete repros and the regression coverage needed before merge.

crates/palamedes/src/transform/server_functions.rs
swernerx: Implemented in ec822385826c3c43ba556e341bf80dd1adf7d1cd.

The resolver now runs OXC semantic analysis only for module-level `"use server"` files, maps eligible local async handlers by `SymbolId`, resolves each wrapper argument through its `ReferenceId`, and excludes any candidate binding with writes. New regressions cover the class static-block shadow and reassigned function declaration; direct/nested wrappers, aliases, imported/non-async cases, syntax/source-map validity, and deterministic output remain covered.

Validation: `cargo +1.95 test --workspace --locked`; `cargo +1.95 clippy --workspace --all-targets --locked -- -D warnings`; `pnpm format:check`; `pnpm lint`; `RUSTUP_TOOLCHAIN=1.95 pnpm build`, `test`, and `check-types`.

swernerx (commented):
Verdict: clear

The semantic correction fully addresses the prior binding-safety finding. Eligible handlers are keyed by OXC `SymbolId`, wrapper arguments resolve through their `ReferenceId`, and bindings with writes are excluded. Both original repros now stay unchanged, while immutable declarations and const functions/arrows, nested wrappers, export aliases, reference expressions, deduplication, and existing exported-initializer behavior still work.

The full workspace tests, formatting, workspace clippy, and an independent syntax/source-map/determinism edge-case harness all pass. Semantic analysis is gated behind the module-level `"use server"` path, the aligned `oxc_semantic` dependency and lockfile additions are minimal, the shipped-binary-size check passed, and all seven CI jobs are green at `ec822385826c3c43ba556e341bf80dd1adf7d1cd`.

PR #602, opened 2026-08-07
fix(next): initialize client fragments before module body

### LINKED ISSUE #588: fix(next): client bootstrap is appended after the module body — module-scope macros evaluate before their messages load
## Problem

`packages/next-plugin/palamedes-loader.cjs:224-226` appends the client message bootstrap (`clientMessageBootstrap`, `:129-148`: read `document.documentElement.lang`, `await Promise.all([...loaders])`, `initializeClientI18n`) **after** the original module body. Statements execute in order, so all original top-level statements run before i18n is installed. The transform compiles macros to inline `getI18n()._(...)`, so an authored module-scope macro (`export const LABEL = t`...``) throws "No active client i18n instance" (`packages/runtime/src/index.ts:292-299`) — or, if another module initialized first, renders fallback text because its own fragment is not loaded yet.

The option doc at `packages/next-plugin/src/index.ts:171-173` claims each module "loads only its own compiled fragment for the document locale **before it evaluates**" — true only for importers (via TLA), not for the module itself.

## Impact

In a split app, the first-evaluated module using a module-scope macro throws during hydration (async-module rejection → all importers fail, interactivity lost); the doc claim hides the constraint.

## Direction

Emit the bootstrap before the body (imports hoist, so it is legal; needs a source-map line offset or wrapper-module approach), or detect module-scope macro output in the transform and warn/fail at build time — and fix the doc comment either way.

swernerx:
Closes #588

## Summary

- bootstrap graph-split client modules before their authored body while retaining directive prologues
- shift source-map mappings so transformed source diagnostics remain aligned
- add executable regressions for module-scope compiled calls, importer waiting/rejection, active-locale fragments, and source-map lines

## Validation

- `RUSTUP_TOOLCHAIN=1.95 pnpm agent:check`
- `pnpm --filter @palamedes/next-plugin test`

### DIFF
--- a/packages/next-plugin/palamedes-loader.cjs
+++ b/packages/next-plugin/palamedes-loader.cjs
@@ -239,49 +239,144 @@ function directivePrologueEnd(code) {
   return end
 }
 
-function offsetSourceMapForInsertion(sourceMap, insertionOffset, insertion) {
-  if (!sourceMap || typeof sourceMap !== "object" || typeof sourceMap.mappings !== "string") {
-    return sourceMap
+function generatedPositionForText(text) {
+  const line = text.split("\n").length - 1
+  return { line, column: text.length - text.lastIndexOf("\n") - 1 }
+}
+
+function compareGeneratedPositions(left, right) {
+  if (left.line !== right.line) {
+    return left.line - right.line
   }
-  if (sourceMap.mappings === "" || insertion.length === 0) {
+  return left.column - right.column
+}
+
+function insertionMetrics(insertion) {
+  const position = generatedPositionForText(insertion)
+  return { addedLines: position.line, finalLineLength: position.column }
+}
+
+function shiftedGeneratedPosition(position, insertionPosition, insertion, metrics) {
+  if (compareGeneratedPositions(position, insertionPosition) < 0) {
+    return position
+  }
+  if (metrics.addedLines === 0) {
+    return { line: position.line, column: position.column + insertion.length }
+  }
+  return {
+    line: position.line + metrics.addedLines,
+    column:
+      position.line === insertionPosition.line
+        ? metrics.finalLineLength + position.column - insertionPosition.column
+        : position.column,
+  }
+}
+
+function isGeneratedPosition(position) {
+  return (
+    position &&
+    typeof position === "object" &&
+    Number.isInteger(position.line) &&
+    position.line >= 0 &&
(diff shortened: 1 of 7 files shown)


packages/next-plugin/palamedes-loader.cjs
```diff
@@ -148,6 +149,152 @@ function clientMessageBootstrap(config, sourcePath, compiledIds) {
...
+      continue
+    }
+    break
+  }
+
+  return end
+}
+
+function offsetSourceMapForInsertion(sourceMap, insertionOffset, insertion) {
+  if (!sourceMap || typeof sourceMap !== "object" || typeof sourceMap.mappings !== "string") {
```
swernerx: P2: Indexed source maps have `sections` instead of a top-level `mappings` string, so this guard returns them unchanged after the bootstrap adds generated lines. Any upstream loader that hands us a v3 indexed map leaves section offsets anchored to the old output, moving diagnostics and stack locations. Please rebase the affected section offsets (or flatten the map first) and add an indexed-map regression.

swernerx (commented):
Verdict: changes requested

The bootstrap now runs before authored module statements while preserving the directive prologue, and the async rejection path is owned by the module. The focused loader suite passed, and all seven checks are green. One source-map compatibility gap remains inline.

packages/next-plugin/palamedes-loader.cjs
swernerx: Implemented in `feb5045e2766e6a3205d280d17b707d21862ee94`: indexed maps now rebase sections at/after the bootstrap and shift nested mappings in the active section, without mutating the input map. Added line and same-line-column regressions, including multiple and empty sections.

Validation passed: `pnpm --filter @palamedes/next-plugin exec vitest run --globals src/palamedes-loader.test.ts`, `pnpm --filter @palamedes/next-plugin test`, `pnpm format:check`, `pnpm lint`, and `RUSTUP_TOOLCHAIN=1.95 pnpm check-types`.

swernerx (commented):
Verdict: clear

The indexed-map follow-up fixes the original P2: sections before, at, and spanning the insertion now rebase correctly through flat and nested maps without mutating caller input. The focused loader suite and an independent position oracle passed the requested boundary, metadata, and directive-order cases, and all seven exact-head checks are green.

PR #604, opened 2026-08-07
fix(core): reject invalid plural ICU patches

### LINKED ISSUE #569: fix(core): invalid ICU in a plural-branch patch aborts the whole batch with a context-free error instead of a structured rejection
## Problem

`validate_patch_translation` (`crates/palamedes/src/translation_candidates.rs:855-897`) checks only shape/keys/emptiness and never parses branch text. The branch is first parsed in `patch_changes_candidate` → `render_translation_value` → `parse_plural_branch` (`:971-1020`), whose error propagates via `?` (`:503`, `:513`) out of `apply_translation_patches` as a top-level `Err`.

Reproduced: a plural branch value `"# Dateien }"` (stray closing brace — plausible provider/LLM output) produces a hard `Err: Unexpected trailing input at line 1, column 36` — line/column refer to the internally synthesized wrapper string, with no patch id, no catalog path, no outcomes report.

This contradicts the documented contract (`docs/translation-candidate-patches.md:36-40`): validation problems "are returned as structured diagnostics".

## Impact

One bad branch in a 200-patch batch turns a recoverable per-entry rejection into an opaque exception; the caller cannot tell which patch failed. Related edge worth fixing in the same pass: multi-file batches are per-file atomic by design, but if `atomic_replace_catalog` fails on a later file (`:561-563`), earlier files are already replaced and the outcomes report is lost in the thrown error.

## Direction

Parse each plural branch during the validation phase and emit a structured per-patch diagnostic (e.g. `translation.invalid_icu` carrying the patch id); keep hard errors for I/O only — and return outcomes alongside per-file write errors.

*Filed from the 2026-08-07 repository audit (reproduced by execution in two independent reviews).*

swernerx:
## Summary

- Validate every plural patch branch before catalog mutation and return patch-scoped `translation.invalid_icu` diagnostics.
- Preserve completed per-file outcomes in a hard write error when a later replacement fails.
- Document per-file atomicity and cover plural, selectordinal, batch isolation, and write-failure behavior.

## Validation

- `cargo +1.97.1 test --workspace --locked`
- `cargo +1.97.1 clippy --workspace --all-targets --locked -- -D warnings`
- `cargo +1.97.1 fmt --all --check`
- `pnpm format:check`
- `RUSTUP_TOOLCHAIN=1.97.1 pnpm check:native-types`

Closes #569

### DIFF
--- a/crates/palamedes/src/error.rs
+++ b/crates/palamedes/src/error.rs
@@ -2,6 +2,8 @@ use std::path::PathBuf;
 
 use thiserror::Error;
 
+use crate::TranslationPatchResult;
+
 /// Shared result type for the Rust core.
 pub type PalamedesResult<T> = Result<T, PalamedesError>;
 
@@ -34,6 +36,17 @@ pub enum PalamedesError {
         #[source]
         /// Underlying filesystem error.
         source: std::io::Error,
+    },
+    /// Replacing a translation catalog failed after one or more catalog outcomes completed.
+    #[error(
+        "Failed to replace a translation catalog; inspect translation_patch_result() for completed outcomes: {source}"
+    )]
+    TranslationPatchWrite {
+        /// Completed per-file outcomes before the failing catalog replacement.
+        result: TranslationPatchResult,
+        #[source]
+        /// Underlying catalog replacement failure.
+        source: Box<PalamedesError>,
     },
     /// The worker pool for parallel extraction could not be created.
     #[error("Could not create the extraction worker pool: {message}")]
@@ -224,4 +237,15 @@ pub enum PalamedesError {
         /// Source location of the nested macro.
         location: String,
     },
+}
+
+impl PalamedesError {
+    /// Returns completed translation-patch outcomes retained with a write failure.
+    #[must_use]
+    pub fn translation_patch_result(&self) -> Option<&TranslationPatchResult> {
+        match self {
+            Self::TranslationPatchWrite { result, .. } => Some(result),
+            _ => None,
+        }
+    }
 }
--- a/crates/palamedes/src/translation_candidates.rs
+++ b/crates/palamedes/src/translation_candidates.rs
@@ -373,6 +373,16 @@ pub fn list_translation_candidates(
 pub fn apply_translation_patches(
     request: TranslationPatchRequest,
 ) -> PalamedesResult<TranslationPatchResult> {
+    apply_translation_patches_with_replacement(request, atomic_replace_catalog)
+}
+
+fn apply_translation_patches_with_replacement<F>(
+    request: TranslationPatchRequest,
+    mut replace_catalog: F,
+) -> PalamedesResult<TranslationPatchResult>
+where
+    F: FnMut(&PreparedCatalog, &str, Option<&PoOutputOptions>) -> PalamedesResult<()>,
+{
     let requested_count = request.patches.len();
     if request.patches.is_empty() {
         return Ok(TranslationPatchResult {
@@ -579,7 +589,7 @@ pub fn apply_translation_patches(
     for catalog in &prepared {
         if catalog.changed {
             if let Err(source) =
-                atomic_replace_catalog(catalog, &request.config.source_locale, request.po.as_ref())
+                replace_catalog(catalog, &request.config.source_locale, request.po.as_ref())
             {
                 return Err(PalamedesError::TranslationPatchWrite {
                     result: completed_translation_patch_result(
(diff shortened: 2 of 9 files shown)


crates/palamedes/src/error.rs
```diff
@@ -35,6 +37,17 @@ pub enum PalamedesError {
         /// Underlying filesystem error.
         source: std::io::Error,
     },
+    /// Replacing a translation catalog failed after one or more catalog outcomes completed.
+    #[error(
+        "Failed to replace a translation catalog; inspect translation_patch_result() for completed outcomes: {source}"
```
swernerx, line 42: [P1] Preserve completed outcomes at the Node boundary

`applyTranslationPatches` still routes this variant through `to_napi_error`, which consumes the owned `TranslationPatchResult` and emits only this display string. JS callers therefore still lose the completed per-file outcomes that issue #569 asks the error to retain, and the exception tells them to call a Rust-only method. Map `TranslationPatchWrite` separately to a documented N-API error/result shape containing the converted report, with a core-node test for a later-file write failure.

crates/palamedes/src/translation_candidates.rs
```diff
@@ -1488,6 +1564,217 @@ mod tests {
...
+        );
+        assert!(result.diagnostics.iter().any(|diagnostic| {
+            diagnostic.code == "translation.invalid_icu"
+                && diagnostic.id.as_ref() == Some(&ordinal.id)
+                && diagnostic.message.contains("translation.values.other")
+        }));
+        assert_eq!(fs::read_to_string(&path).unwrap(), before);
+    }
+
+    #[cfg(unix)]
```
swernerx: [P2] Make the failure proof portable

This gate removes the only `TranslationPatchWrite` behavior test on Windows, while chmod 0555 is not guaranteed to deny writes for privileged Unix test users. Use a deterministic failure-injection seam or another cross-platform replacement failure, keep the partial-outcome assertions shared, and assert any platform-specific source behavior explicitly.

swernerx (commented):
Verdict: changes requested

The core validation path is sound: every plural/selectordinal branch is parsed before mutation, diagnostics remain patch-scoped and deterministic, and the per-file outcome accounting is internally consistent. Exact-head translation-candidate tests pass 9/9 and exact-head CI is 7/7 green.

Two boundary gaps still block approval; details are inline.

crates/palamedes/src/error.rs
swernerx: Implemented in 74633013396c5578299f17410a6c75f19ec5f2b6. The N-API binding now throws a hard Error with stable code ERR_PALAMEDES_TRANSLATION_PATCH_WRITE, a catalog-write cause (ERR_PALAMEDES_CATALOG_WRITE), and a converted partial TranslationPatchResult on error.report; the TypeScript wrapper exports the corresponding type, constants, and guard while normalizing report statuses. Added an addon integration test for a deterministic later-file replacement failure that verifies the error, report, and both catalog contents. Validation: cargo +1.97.1 test --workspace --locked; cargo +1.97.1 clippy --workspace --all-targets --locked -- -D warnings; cargo +1.97.1 fmt --all --check; pnpm test; pnpm check-types; pnpm lint; pnpm build:site.

crates/palamedes/src/translation_candidates.rs
swernerx: Implemented in 74633013396c5578299f17410a6c75f19ec5f2b6. Replaced the Unix chmod proof with a private closure-scoped replacement seam, so the core test runs on every platform and deterministically injects a WriteFile source for the second catalog. It preserves the shared partial-outcome assertions, verifies the underlying source kind/message, and verifies the first catalog changed while the failed source catalog remained unchanged. The Node integration uses a doc-hidden test-support feature only while building its test addon; no default production addon exposes the failure injector or uses global state. Validation: cargo +1.97.1 test --workspace --locked; cargo +1.97.1 clippy --workspace --all-targets --locked -- -D warnings; cargo +1.97.1 fmt --all --check; pnpm test; pnpm check-types; pnpm lint; pnpm build:site.

crates/palamedes/src/error.rs
swernerx: Follow-up: the current fixing head is a461f5001007f4c51ed01c4ee078198abeb90e7f. The only follow-up change is the integration test timeout required for a cold feature-addon build in Node 22 CI; the N-API error/report API described above is unchanged. Focused core-node test and format check pass locally; the replacement seven-check matrix is pending.

crates/palamedes/src/translation_candidates.rs
swernerx: Follow-up: the current fixing head is a461f5001007f4c51ed01c4ee078198abeb90e7f. Node 22 CI confirmed that the cross-platform feature-addon integration test exceeded Vitest’s 5s default only because its cold Cargo build took 6.5s. The test now has an explicit 30s limit; its deterministic failure injection and all source/report assertions are unchanged. Focused core-node test and format check pass locally; the replacement seven-check matrix is pending.

swernerx (commented):
Verdict: clear

Both prior findings are addressed at exact head `a461f5001007f4c51ed01c4ee078198abeb90e7f`. The Node boundary preserves a hard Error with a stable code, catalog-write cause, and normalized partial report; the deterministic feature-gated addon test covers the cross-platform later-file failure without exposing test support in the default addon. Focused exact-head core and addon checks pass, and all seven exact-head CI checks are green.

PR #607, opened 2026-08-07
fix(next): degrade failed client catalog fragments

### LINKED ISSUE #591: feat(next): one failed catalog-fragment request kills hydration of the whole client graph
## Problem

The injected client bootstrap does a top-level `await Promise.all([...])` with no error handling (`packages/next-plugin/palamedes-loader.cjs:136-140`). A rejected fragment import rejects the async module, and every transitive importer — ultimately the entry — fails to evaluate.

Commit 6e86e9d (`allowedDevOrigins` for the dev test) was a legitimate Next 16 dev-origin accommodation, but it demonstrates the amplification: SSR HTML renders, then a single failed `.po` chunk leaves the page permanently non-interactive.

## Impact

In production, one flaky/blocked chunk request (network hiccup, aggressive ad-blocker, stale CDN after deploy) disables **all** client interactivity rather than degrading one module's translations.

## Direction

Catch fragment-loader rejection per catalog, log, and continue with source fallbacks (at minimum in production builds), or retry the import once. This is a deliberate loud-vs-degrade trade-off — worth an explicit decision either way.

swernerx:
## Summary

- make production graph-split client bootstraps handle catalog fragment imports independently
- safely log and skip only a rejected fragment while loading successful sibling fragments and completing hydration
- retain fail-fast behavior in development; core/runtime bootstrap failures remain fatal

## Behavior

Production uses a deliberate no-immediate-retry degradation policy. A rejected catalog fragment is logged with safely serialized module-path context and the locale, then skipped without allowing a faulty logger to reject the module. Palamedes avoids an immediate, unbacked-off retry because likely deterministic CDN, ad-blocker, or stale-deploy failures generally need backoff or a later navigation rather than another request in the same bootstrap turn. Set `keepSourceFallbacks: true` when readable source text is required after a production fragment is skipped.

## Validation

- `vitest run packages/next-plugin/src/palamedes-loader.test.ts packages/next-plugin/src/index.test.ts` — 39 passed
- repository-wide `oxfmt --check` and `oxlint --deny-warnings`
- `node --check packages/next-plugin/palamedes-loader.cjs`
- `git diff --check`

A direct package typecheck cannot run authoritatively from this fresh worktree: dependencies are reused read-only from the saved checkout, whose workspace symlinks resolve against a different source tree and produce unrelated existing errors for `@palamedes/next-plugin/server-function-entry`, `loadRegisteredMessages`, and `createI18n`. Exact-head CI performs the authoritative typecheck.

Closes #591

### DIFF
--- a/packages/next-plugin/README.md
+++ b/packages/next-plugin/README.md
@@ -145,6 +145,16 @@ invalidate their affected subsets under Turbopack and webpack; Next may apply
 Fast Refresh or fall back to a full document reload at an async-module
 boundary. A document reload is the supported fallback.
 
+Each production fragment import is attempted once. If one rejects, Palamedes
+logs the failure with the module path and locale, skips only that fragment, and
+continues hydrating the client graph (including any other fragments that did
+load). It deliberately does not immediately retry: browsers commonly retain a
+failed dynamic-import result, so a retry adds noise without making the graph
+more reliable. Development remains fail-fast so catalog wiring failures stay
+visible while editing. In a production build, set `keepSourceFallbacks: true`
+if readable source text is required when a skipped fragment has no loaded
+translation; the compact default does not retain source messages.
+
 `messageSplitting` currently supports PO catalogs and defaults to `false` for
 compatibility. Keep using `createClientCatalogBoundary()` from
 `@palamedes/react/client` when an app needs a complete active-locale catalog or
--- a/packages/next-plugin/README.md
+++ b/packages/next-plugin/README.md
@@ -148,10 +148,11 @@ boundary. A document reload is the supported fallback.
 Each production fragment import is attempted once. If one rejects, Palamedes
 logs the failure with the module path and locale, skips only that fragment, and
 continues hydrating the client graph (including any other fragments that did
-load). It deliberately does not immediately retry: browsers commonly retain a
-failed dynamic-import result, so a retry adds noise without making the graph
-more reliable. Development remains fail-fast so catalog wiring failures stay
-visible while editing. In a production build, set `keepSourceFallbacks: true`
+load). It deliberately avoids an immediate, unbacked-off retry: deterministic
+CDN, ad-blocker, and stale-deploy failures are unlikely to improve in the same
+turn, while a retry can add a request without restoring the graph. Development
+remains fail-fast so catalog wiring failures stay visible while editing. In a
+production build, set `keepSourceFallbacks: true`
 if readable source text is required when a skipped fragment has no loaded
 translation; the compact default does not retain source messages.
 
--- a/packages/next-plugin/palamedes-loader.cjs
+++ b/packages/next-plugin/palamedes-loader.cjs
@@ -106,7 +106,7 @@ function selectedMessageImports(config, sourcePath, compiledIds) {
   )
 }
 
-function clientMessageBootstrap(config, sourcePath, compiledIds) {
+function clientMessageBootstrap(config, sourcePath, compiledIds, fragmentFailureMode) {
   const importsByCatalog = selectedMessageImports(config, sourcePath, compiledIds)
   if (!importsByCatalog) {
     return null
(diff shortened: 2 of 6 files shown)


packages/next-plugin/palamedes-loader.cjs
```diff
@@ -127,26 +127,42 @@ function clientMessageBootstrap(config, sourcePath, compiledIds) {
...
-  )
+    `);\n`
+  const fragmentImports =
+    fragmentFailureMode === "degrade"
+      ? `const ${identifier}_fragments = await Promise.all(${identifier}_activeLoaders.map(async (load) => {\n` +
+        `  try {\n` +
+        `    return await load();\n` +
+        `  } catch (error) {\n` +
+        `    console.error(\n` +
+        `      \`Palamedes client graph message splitting failed to load a catalog fragment for ${modulePath} (\${${identifier}_locale}). Continuing without that fragment.\`,\n` +
```
swernerx: The recovery path can still reject here if `console.error` throws: that exception escapes the per-fragment task before `return null`, so `Promise.all` rejects and hydration is lost again. The generated template also inserts `modulePath` as raw template-literal source; a valid filename containing a backtick produces invalid JavaScript, while `${...}` in a filename is evaluated when this log runs. Please emit the diagnostic from JSON-escaped literals, guard the logger so it cannot undo degradation, and cover a throwing logger in the behavioral regression.

packages/next-plugin/README.md
```diff
@@ -145,6 +145,16 @@ invalidate their affected subsets under Turbopack and webpack; Next may apply
 Fast Refresh or fall back to a full document reload at an async-module
 boundary. A document reload is the supported fallback.
 
+Each production fragment import is attempted once. If one rejects, Palamedes
+logs the failure with the module path and locale, skips only that fragment, and
+continues hydrating the client graph (including any other fragments that did
+load). It deliberately does not immediately retry: browsers commonly retain a
```
swernerx: The once-only policy is reasonable, but this rationale overstates browser behavior. Failed loads/links may be retried under module-loading semantics, and the supported webpack runtime explicitly resets a failed chunk to unloaded, so a second `import()` can issue another request; only evaluation failures are reliably cached. Please document the actual policy trade-off (for example, avoiding an immediate, unbacked-off retry of deterministic CDN/ad-block/stale-deploy failures), update the same claim in the PR description, and qualify the unconditional fragment-loading statements in `docs/api/next-plugin.md` for production degradation.

swernerx (commented):
Verdict: changes requested

The per-catalog isolation is otherwise well targeted: the regression proves a sibling fragment still loads and the generated async module completes, development stays fail-fast, and core/runtime bootstrap failures remain fatal. Two P2 issues remain in the production diagnostic and documentation paths; details are inline.

packages/next-plugin/palamedes-loader.cjs
swernerx: Fixed in 82ca1db180470c2c31b14b5ecf80abda54fb94c3.

The generated diagnostic now uses JSON-serialized static prefix/suffix literals with the locale passed as a separate console argument, so adversarial paths cannot break or execute generated code. Logging is wrapped so a throwing or unavailable logger cannot undo fragment degradation. The regression covers an adversarial backtick/interpolation path, a throwing logger, sibling-fragment loading, and async module completion.

Validation: 39 focused Next-plugin tests passed; repository-wide oxfmt and oxlint passed; all seven exact-head CI checks passed.

packages/next-plugin/README.md
swernerx: Fixed in 82ca1db180470c2c31b14b5ecf80abda54fb94c3.

The package README, Next API documentation, and PR description now describe the actual policy trade-off: production avoids an immediate, unbacked-off retry of likely deterministic CDN, ad-blocker, or stale-deploy failures; a later navigation or backoff is the recovery path. The API docs now explicitly qualify production fragment rejection as logged-and-skipped, with fallback behavior documented.

Validation: 39 focused Next-plugin tests passed; repository-wide oxfmt and oxlint passed; all seven exact-head CI checks passed.

swernerx (commented):
Verdict: clear

The fixing commit closes both prior P2s. Generated diagnostics now use JSON-serialized static literals with locale and error passed separately, and logging failures are contained so they cannot reject production recovery. The behavioral coverage proves adversarial paths remain inert, a throwing logger cannot block sibling fragments or async-module completion, development remains fail-fast, and core/runtime bootstrap failures remain fatal. The README, Next API reference, and PR description now accurately document the no-immediate-unbacked-off-retry policy, logged-and-skipped production behavior, and source-fallback trade-off. No P0-P2 finding remains at exact head 82ca1db180470c2c31b14b5ecf80abda54fb94c3; all seven checks are successful.

PR #608, opened 2026-08-07
fix(next): harden split-catalog dev invalidation

### LINKED ISSUE #590: test(next): dev invalidation silently degrades without addDependency; fallback/config edits are the untested path; deleted-module registrations linger
Three dev-mode robustness gaps in the split-catalog pipeline:

## 1. Silent degradation when the loader host lacks `addDependency`

`palamedes-po-loader.cjs:92-99` and `palamedes-loader.cjs:218-220` guard with `typeof this.addDependency === "function"` and silently skip registration when absent. If a bundler shim ignores the calls, a dev edit to the source-locale catalog or the config leaves every split sidecar stale until restart — silently. Direction: one-time warning when `addDependency` is unavailable while splitting is active.

## 2. The only CI dev-invalidation test edits `de.po` — the path that works even without `addDependency`

`examples/nextjs-cookie/scripts/message-splitting-development.test.mjs:82-94` (Turbopack only) edits the sidecar's own *resource*. Editing `en.po` (source-fallback content compiled into `de` sidecars) or the config relies entirely on `addDependency` being honored and has no test. Direction: extend the dev test to edit `en.po` (asserting a de page whose entry is missing in de.po updates) and to touch the config; also note webpack dev with TLA client modules has no HMR test.

## 3. Registrations of deleted/renamed server modules are never removed in dev

`packages/runtime/src/index.ts:144-152` only sets/replaces by stable key; nothing unregisters when a module leaves the graph, and `loadRegisteredMessages` iterates every state per server-function request. Long dev sessions accumulate stale loaders; removed translations linger until restart. Production is unaffected. Direction: stamp registrations with a build generation and skip older ones, or document the limitation next to the "superset" note in the README.

swernerx:
## Summary

- Emit one development-only warning per loader compilation when message splitting cannot register dependencies with `addDependency()`.
- Watch source-locale fallback catalogs and configuration changes in the real Turbopack development regression.
- Replace server-side lazy loader registrations by source-module group and release webpack registrations on HMR disposal.

## Behavior and limitations

Supported hosts continue to use dependency registration and production remains warning-free. A host without `addDependency()` receives one actionable development warning and requires a restart after fallback/config edits.

The development regression validates Turbopack's supported document-reload fallback. Webpack top-level-await client output remains production-build-covered; this PR does not claim an equivalent HMR contract. Webpack server modules now release exact registrations on disposal; Turbopack does not expose a compatible server-module disposal hook, so a module removed from its graph can remain registered until the development server restarts.

## Validation

- `RUSTUP_TOOLCHAIN=1.97.1 pnpm build`
- `RUSTUP_TOOLCHAIN=1.97.1 pnpm check-types`
- `pnpm --filter @palamedes/next-plugin test`
- `pnpm --filter @palamedes/runtime test`
- `pnpm --filter @palamedes/example-nextjs-cookie test:message-splitting:development`
- `pnpm lint`
- `pnpm format:check`

Closes #590

### DIFF
--- a/packages/next-plugin/palamedes-loader.cjs
+++ b/packages/next-plugin/palamedes-loader.cjs
@@ -535,6 +535,7 @@ module.exports = function palamedesLoader(source, inputSourceMap) {
     }
     if (
       serverMessageSplitting &&
+      result.compiledIds.length > 0 &&
       !registration?.matchesCatalog &&
       typeof this.emitWarning === "function"
     ) {
--- a/packages/next-plugin/palamedes-loader.cjs
+++ b/packages/next-plugin/palamedes-loader.cjs
@@ -7,6 +7,7 @@ const { decode, encode } = require("@jridgewell/sourcemap-codec")
 const { loadPalamedesConfigSync } = require("@palamedes/config")
 const { transformPalamedesMacros } = require("@palamedes/transform")
 const picomatch = require("picomatch")
+const { warnMissingAddDependency } = require("./palamedes-dev-warning.cjs")
 
 const SELECTED_MESSAGES_QUERY = "palamedes-selected"
 const configCache = new Map()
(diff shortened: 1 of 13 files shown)


packages/next-plugin/palamedes-loader.cjs
```diff
@@ -459,24 +479,43 @@ module.exports = function palamedesLoader(source, inputSourceMap) {
...
         code = output.code
         sourceMap = offsetSourceMapForInsertion(sourceMap, output.insertion, output.insertion.value)
       } else {
-        code += registration
+        code += registrationCode
       }
-    } else if (typeof this.emitWarning === "function") {
+    }
+    if (
+      serverMessageSplitting &&
```
swernerx, line 537: The new empty-ID cleanup also reaches this branch. With `serverFunctions` enabled, webpack applies this loader to every included server JS/TS module, so a message-free file outside all catalog scopes emits `uses messages...` once per file on every compilation. I reproduced it with `compiledIds: []`: the loader generated the empty group cleanup and still emitted the warning. That turns a narrow catalog config into false, potentially high-volume diagnostics. Keep the cleanup registration, but gate this warning on `result.compiledIds.length > 0` and cover the no-message/outside-catalog case.

swernerx (commented):
Verdict: changes requested

The fallback/config invalidation and registration ownership paths are otherwise in good shape, and the exact-head checks are green. One P2 remains in the development cleanup diagnostic; details inline.

packages/next-plugin/palamedes-loader.cjs
swernerx: Fixed in 28238ce9de5998c365c94076ec8976b1fdd142f3. The outside-catalog warning now requires at least one compiled message ID, while development still emits the empty registration group for message-free modules that leave catalog scope.

Validation: `pnpm --filter @palamedes/next-plugin test` (48 passed), `pnpm --filter @palamedes/runtime test` (41 passed), `RUSTUP_TOOLCHAIN=1.97.1 pnpm check-types`, `pnpm lint`, and `pnpm format:check`.

swernerx (commented):
Verdict: clear

The empty-ID cleanup now keeps the source-module registration group without emitting the outside-catalog warning. Actual compiled IDs still warn, and the missing-addDependency diagnostic remains independent. The regression exercises the shipped loader path and fails against the prior head. The surrounding invalidation, HMR ownership, production behavior, and documented Turbopack limitation remain intact.

PR #609, opened 2026-08-07
perf(next): cache selected sidecar catalog snapshots

### LINKED ISSUE #589: perf(next): every split sidecar compile re-parses all catalogs — O(modules × locales × catalog size), amplified on dev edits
## Problem

`packages/next-plugin/palamedes-po-loader.cjs:76-80` calls `compileCatalogArtifactSelected` once per sidecar module (one per message-bearing source module × locale). Each call runs `prepare_compilation` → `load_catalogs` (`crates/palamedes/src/catalog_artifact/load.rs:13-56`), which reads and parses the primary **and every fallback-chain** .po file and rebuilds the full `CompiledCatalogIdIndex` (`catalog_artifact/mod.rs:101-106`) — no cache at any layer (the JS `configCache` caches only the config object).

An edit to `de.po` invalidates every de sidecar, so one locale-file save in dev triggers N full catalog parses + N id-index builds for N message-bearing modules. A 500-module, 5-locale app performs 2500 selected compiles on a cold build, each re-parsing the whole catalog and its fallbacks.

## Impact

Cold-build and dev-rebuild latency scales superlinearly with app size. Correctness is unaffected.

## Direction

Cache `NormalizedParsedCatalog` and `CompiledCatalogIdIndex` per (path, mtime/content hash) in the native layer, or memoize per compilation in the loader process, and reuse across sidecar compiles.

swernerx:
Closes #589

## Summary

- Add a bounded, content-addressed cache for selected catalog compilation snapshots in the native Node process.
- Reuse normalized primary/fallback catalogs and the compiled ID index across sidecar selections while leaving generated modules, diagnostics, fallback behavior, and public Node APIs unchanged.
- Read and SHA-256 every dependency before each cache lookup. Content replacement with unchanged mtimes, newly present/missing fallbacks, parse/read failures, source-locale changes, and fallback/config changes cannot return a stale snapshot.
- Coalesce identical concurrent snapshot construction per cache key while allowing unrelated catalog sets to proceed independently. The N-API cache holds at most 64 immutable snapshots.

## Evidence

- New deterministic scaling regression: 64 selected compiles reuse two catalog parses and one ID-index build.
- New equal-mtime replacement regression confirms changed content is recompiled and malformed content still surfaces an error.
- `cargo +1.95 test -p palamedes --lib` (288 passed)
- `cargo +1.95 clippy -p palamedes -p palamedes-node -- -D warnings`
- Next loader Vitest could not run in this fresh worktree because its incomplete dependency store attempted registry downloads and sandbox DNS cannot resolve the registry.

### DIFF
--- a/crates/palamedes/src/catalog_artifact/cache.rs
+++ b/crates/palamedes/src/catalog_artifact/cache.rs
@@ -1,5 +1,5 @@
 use std::collections::HashMap;
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Condvar, Mutex, MutexGuard, PoisonError};
 
 use ferrocat::{CompiledCatalogIdIndex, CompiledKeyStrategy};
 
@@ -26,12 +26,20 @@ pub struct CatalogCompilationCache {
     state: Mutex<CacheState>,
     #[cfg(test)]
     statistics: Mutex<CacheStatistics>,
+    #[cfg(test)]
+    before_build: Mutex<Option<Arc<dyn Fn(&str) + Send + Sync>>>,
 }
 
 struct CacheState {
-    entries: HashMap<CompilationCacheKey, Arc<Mutex<Option<Arc<CachedCompilation>>>>>,
+    ready: HashMap<CompilationCacheKey, Arc<CachedCompilation>>,
+    in_flight: HashMap<CompilationCacheKey, Arc<InFlightBuild>>,
     clock: u64,
     last_used: HashMap<CompilationCacheKey, u64>,
+}
+
+struct InFlightBuild {
+    finished: Mutex<bool>,
+    wake: Condvar,
 }
 
 struct CachedCompilation {
@@ -73,12 +81,15 @@ impl CatalogCompilationCache {
         Self {
             capacity,
             state: Mutex::new(CacheState {
-                entries: HashMap::new(),
+                ready: HashMap::new(),
+                in_flight: HashMap::new(),
                 clock: 0,
                 last_used: HashMap::new(),
             }),
             #[cfg(test)]
             statistics: Mutex::new(CacheStatistics::default()),
+            #[cfg(test)]
+            before_build: Mutex::new(None),
         }
     }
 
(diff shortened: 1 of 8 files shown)


crates/palamedes/src/catalog_artifact/cache.rs
```diff
@@ -0,0 +1,204 @@
...
+        if self.capacity == 0 {
+            let cached = self.build(snapshot)?;
+            return compile_selected_prepared(&cached.prepared, &cached.compiled_id_index, request);
+        }
+
+        let entry = self.entry(key);
+        let compiled = {
+            let mut entry = entry.lock().expect("catalog cache entry lock poisoned");
+            if entry.is_none() {
+                *entry = Some(Arc::new(self.build(snapshot)?));
```
swernerx: Failed builds are retained here as `None`: the `?` returns while this key remains in both `entries` and `last_used`, so repeated malformed snapshots can consume the 64 slots and evict successful hot snapshots. The same representation lets capacity eviction remove an in-flight `None`; a later request for that key creates another mutex and repeats parse/index work instead of coalescing. Please separate in-flight state from the bounded ready-entry LRU (or otherwise remove matching failed placeholders and never evict active builds), with failure and capacity-pressure concurrency coverage.

crates/palamedes/src/catalog_artifact/tests.rs
```diff
@@ -590,6 +591,98 @@ msgstr "Hallo"
...
+        &locale_dir,
+        "de",
+        &[("Hello", "Hallo"), ("Goodbye", "Auf Wiedersehen")],
+    );
+    let request = selected_request(&fixture, &locale_dir, "Hello");
+    let cache = CatalogCompilationCache::new(8);
+
+    // This models many module sidecars selecting different subsets from one
+    // locale. Parsing/index construction scales with catalog snapshots, not
+    // with the number of selected module compiles.
```
swernerx: This loop reuses one request and one `compiled_ids` set, so it would still pass if selections were part of the cache key—the exact modules×sidecars amplification the issue targets. `parses` is also inferred from `snapshot.sources.len()` after a successful build rather than counted at the parser call. Please vary all 64 selection sets and instrument the actual parse and index-construction sites so this regression fails for per-selection rebuilds.

swernerx (commented):
Verdict: changes requested

The content snapshot boundary is otherwise sound: it hashes the same bytes it parses, preserves per-request selection and pseudolocalization, and my release probe measured an 11.0× cached/uncached speedup on two 10,000-message catalogs. Two P2 items remain in the inline comments: cache entry lifecycle under failure/capacity pressure, and a regression that does not yet model distinct sidecar selections or count the actual heavy calls.

crates/palamedes/src/catalog_artifact/cache.rs
swernerx: Fixed in 65e201ba10e46d51581eb30f08cbdb54d3a9d7f3.

Ready snapshots and in-flight builds are now separate: only successful immutable snapshots participate in the bounded LRU; failed builds are removed and wake waiters; and an RAII completion guard removes/notifies in-flight entries even during unwinding. The cache recovers poisoned mutexes and does not hold its state lock while parsing. Added deterministic tests for repeated parse failures preserving ready capacity, unrelated-key progress under capacity pressure, identical in-flight build coalescing, and unwind cleanup.

Validated with `cargo +1.95 test -p palamedes --lib` (291 passed), `cargo +1.95 test -p palamedes-node` (5 passed), and Clippy with warnings denied.

crates/palamedes/src/catalog_artifact/tests.rs
swernerx: Fixed in 65e201ba10e46d51581eb30f08cbdb54d3a9d7f3.

The scaling regression now compiles 64 distinct selected IDs (one per sidecar) from a 64-message catalog and asserts exactly two parser calls and one compiled-ID-index construction. Those counters are incremented directly at `parse_catalog` and `CompiledCatalogIdIndex::new_with_policy`, so per-selection cache keys or rebuilding are detected. The same-mtime replacement and malformed-catalog invalidation coverage remains.

Validated with `cargo +1.95 test -p palamedes --lib` (291 passed), `cargo +1.95 test -p palamedes-node` (5 passed), and Clippy with warnings denied.

crates/palamedes/src/catalog_artifact/cache.rs
swernerx: Final fixing head: aa60bd263df87cd10ef50b3e2106472ccbc739dd.

This retains the lifecycle fix in 65e201b and adds the CI-required test-hook type alias. The workspace/all-targets Clippy command now passes, along with 291 core and 5 N-API tests. The thread remains open for your review.

crates/palamedes/src/catalog_artifact/tests.rs
swernerx: Final fixing head: aa60bd263df87cd10ef50b3e2106472ccbc739dd.

This retains the distinct-64-selection regression and direct parser/index construction counters in 65e201b, plus the CI-required test-hook type alias. The workspace/all-targets Clippy command now passes, along with 291 core and 5 N-API tests. The thread remains open for your review.

swernerx (commented):
Verdict: clear

The two prior P2 findings are resolved. Ready snapshots and active builds now have separate lifecycles; successful immutable snapshots alone occupy the 64-entry LRU, while matching failure and unwind cleanup removes and notifies in-flight state without holding the global mutex across heavy work. The regression now exercises 64 distinct sidecar selections and counts the actual parser and compiled-index construction sites. Equal-mtime invalidation, malformed/error recovery, fallback behavior, selected filtering, diagnostics, and the N-API boundary remain intact. The final aa60bd commit only introduces the Clippy-required test-hook type alias on top of 65e201b. Rust 1.95 core/N-API tests, workspace all-target Clippy, formatting, adversarial concurrency/error probes, and the release performance probe pass; all eight live checks are successful.

PR #611, opened 2026-08-07
ci: repair remix next canary

### LINKED ISSUE #586: ci: remix@next canary never builds workspace packages — it has been failing for infrastructure reasons since inception
## Problem

The `remix-next-canary` job (`ci.yml:158-196`) goes checkout → Node → corepack → Rust → `pnpm install` → `pnpm add remix@next` → `pnpm verify:examples:smoke -- --framework remix`. There is no `pnpm build` step, and `verify:examples:smoke` (unlike `verify:examples`) does not build workspace packages — the Remix examples then fail typecheck/module resolution on the gitignored `dist/` (`Cannot find module '@palamedes/core/macro'` etc.).

Verified identical annotations on the current HEAD run and on older green runs — the canary has never tested what it claims to.

## Impact

The advisory canary emits "remix@next canary failed" on every push while testing nothing: zero early warning for upstream Remix beta churn, one wasted runner per CI run, and permanent warning noise that trains everyone to ignore it.

## Direction

Add `pnpm build` (or the `@palamedes/remix...` dependency closure) before the smoke step; optionally move the canary to a schedule so a real upstream failure is visible instead of continue-on-error noise per PR.

swernerx:
## Summary

Repairs the Remix-next canary so it builds the exact `@palamedes/remix` workspace dependency closure before testing every Remix strategy. The canary now runs in an isolated worktree, so `pnpm add remix@next` can update only that disposable checkout.

The new dedicated workflow runs after `main` changes, weekly for upstream churn, and manually; it no longer consumes a runner on every pull request. Setup failures and post-setup upstream compatibility failures are hard failures with distinct step summaries. A CI-enforced static contract prevents removal or reordering of the required build, override, and smoke steps.

## Root cause

The former CI job invoked the smoke verifier without building gitignored workspace `dist/` outputs. Remix example typechecking therefore failed before any meaningful canary compatibility assertion ran.

## Validation

- `pnpm check:remix-next-canary`
- `pnpm format:check`
- `pnpm lint`
- Prettier YAML/package validation and `git diff --check`
- Isolated local canary equivalent: locked install, topological Remix dependency-closure build, `remix@next` override, and all four Remix smoke strategies (typecheck/build/start/runtime assertions)

Closes #586

### DIFF
--- a/.github/workflows/remix-next-canary.yml
+++ b/.github/workflows/remix-next-canary.yml
@@ -1,0 +1,111 @@
+name: Remix next canary
+
+on:
+  push:
+    branches:
+      - main
+  schedule:
+    # Main catches Palamedes regressions; this catches upstream Remix churn.
+    - cron: "23 5 * * 1"
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: remix-next-canary-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  remix-next-canary:
+    name: remix next canary
+    runs-on: ubuntu-24.04
+    timeout-minutes: 30
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v7
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v7
+        with:
+          node-version: 24
+          # This caches pnpm's download store only. remix@next is still resolved
+          # afresh below, so the cache cannot pin or mask the canary version.
+          cache: pnpm
+          cache-dependency-path: pnpm-lock.yaml
+
+      - name: Enable Corepack
+        run: corepack enable
+
+      - name: Set up Rust
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Create isolated canary workspace
+        id: canary-worktree
+        run: |
+          canary_dir="$RUNNER_TEMP/remix-next-canary"
+          git worktree add --detach "$canary_dir" "$GITHUB_SHA"
+          echo "path=$canary_dir" >> "$GITHUB_OUTPUT"
+
+      - name: Install locked dependencies
+        id: install-dependencies
+        working-directory: ${{ steps.canary-worktree.outputs.path }}
+        run: pnpm install --frozen-lockfile
+
+      # Recursive pnpm runs dependency packages before their dependents: the
+      # native platform package precedes its JS wrapper, then the complete
+      # @palamedes/remix workspace closure. The selector keeps unrelated
+      # packages (notably the CLI) out of this compatibility canary.
+      - name: Build workspace packages required by Remix
+        id: build-workspace
+        working-directory: ${{ steps.canary-worktree.outputs.path }}
+        run: pnpm --filter "@palamedes/remix..." -r build
+
+      - name: Override Remix only inside canary workspace
+        id: install-remix-next
+        working-directory: ${{ steps.canary-worktree.outputs.path }}
+        run: pnpm --filter "@palamedes/example-remix-*" add remix@next --save-exact
+
+      - name: Record resolved remix@next version
+        id: remix-next-version
+        working-directory: ${{ steps.canary-worktree.outputs.path }}/examples/remix-cookie
+        run: echo "version=$(node -p \"require('remix/package.json').version\")" >> "$GITHUB_OUTPUT"
+
+      - name: Smoke-test every Remix strategy against remix@next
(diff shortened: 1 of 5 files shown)


.github/workflows/remix-next-canary.yml
```diff
@@ -0,0 +1,111 @@
...
+    runs-on: ubuntu-24.04
+    timeout-minutes: 30
+
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v7
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v7
+        with:
```
swernerx, line 39: [P1] Enable Corepack before asking setup-node to cache pnpm. The Ubuntu 24.04 hosted image does not provide pnpm, and setup-node@v7 resolves this cache by running `pnpm store path --silent` during this step. Because `corepack enable` runs later, the dedicated workflow will fail in “Set up Node.js” before it creates or exercises the canary worktree. Match the established example workflow: set up Node without caching, enable Corepack, then make the cache-enabled setup-node call (or otherwise install pnpm first).

scripts/check-remix-next-canary.mjs
```diff
@@ -0,0 +1,48 @@
+import assert from "node:assert/strict"
+import { readFile } from "node:fs/promises"
+import { EXAMPLE_MATRIX } from "./example-matrix.mjs"
+
+const workflow = await readFile(
+  new URL("../.github/workflows/remix-next-canary.yml", import.meta.url),
+  "utf8"
+)
+
+const requiredSteps = [
```
swernerx: [P2] Make the contract guard cover the isolation behavior it claims to protect. I ran this exact checker against an in-memory workflow with the install/build/override/smoke `working-directory` bindings and the cleanup step removed, and it still passed. That lets CI stay green after `pnpm add` starts mutating the primary checkout or stale worktrees accumulate; the guard also does not reject a `pull_request` trigger being reintroduced. Add semantic assertions for no PR trigger, every mutating step’s temporary working directory, create → install → build → override → version → smoke ordering, and always-run cleanup.

swernerx (commented):
Verdict: changes requested

The dedicated trigger split and disposable-worktree design are directionally right, and the successful PR matrix confirms the shared four-strategy smoke harness. Two findings block the lane from being reliable: setup-node tries to cache pnpm before pnpm exists, and the contract check remains green after core isolation and cleanup behavior is removed.

.github/workflows/remix-next-canary.yml
swernerx: Fixed in dc716321cbed0c8e0ac925eace220795c3b7b50f.

The dedicated lane now uses the established clean-runner bootstrap order: uncached `actions/setup-node@v7` with Node 24, `corepack enable`, then a second cache-enabled `setup-node` step. The semantic guard has adversarial coverage for this ordering and rejects both an initial pnpm cache and a cache step placed before Corepack.

Validated with the 9-probe `pnpm check:remix-next-canary` suite, `pnpm format:check`, `pnpm lint`, `RUSTUP_TOOLCHAIN=1.95-aarch64-apple-darwin pnpm build`, `pnpm check-types`, and `pnpm test`.

scripts/check-remix-next-canary.mjs
swernerx: Fixed in dc716321cbed0c8e0ac925eace220795c3b7b50f.

The checker now parses the workflow as YAML and validates trigger semantics, the Node → Corepack → cache bootstrap, command/worktree bindings for install/build/override/version/smoke, strict create → install → build → override → version → smoke ordering, and always-run worktree cleanup. Its committed Node test suite uses valid parsed/serialized adversarial mutations and rejects: a PR trigger, broken Corepack caching, every missing phase, every broken worktree binding, every adjacent phase reorder, and missing or conditional cleanup.

Validated by all 9 adversarial probes through `pnpm check:remix-next-canary`, plus format, lint, workspace build, typecheck, and package tests.

swernerx (commented):
Verdict: clear

The bootstrap now runs uncached Node setup → Corepack enable → cache-enabled setup-node, so pnpm exists before cache discovery. The parsed-YAML guard and both the committed suite and 34 independent adversarial mutations reject the requested trigger, bootstrap, phase, command, worktree-binding, ordering, cleanup, and serialization failures. The dedicated main/schedule/manual lane remains off pull requests; manual dispatch correctly becomes available once the workflow exists on the default branch. The dependency-closure and four-strategy smoke path are unchanged, and all seven exact-head checks are green.

PR #612, opened 2026-08-07
ci: close verification gaps across release and platforms

### LINKED ISSUE #585: ci: verification gaps — orphaned container tests, JS-only release gate, no package unit tests on Windows/macOS
Three CI verification gaps, related enough to decide together:

## 1. `scripts/container/start-plan.test.mjs` is never executed anywhere

The module split exists explicitly "so the host-binding override logic can be unit-tested" — but no runner references the test: root `pnpm test` filters `./packages/**` only, no workflow, no vitest config includes it. The container host-binding logic is guarded by tests that never run. Direction: wire `node --test scripts/container/*.test.mjs` into the CI full legs or the container publish workflow.

## 2. The publish gate re-validates JS only

`publish.yml` `validate-release` runs release-set/build/published-types/`pnpm test`/check-types — no `cargo test`, no clippy, no lint. Publish triggers on push to main independently of `ci.yml` (no `workflow_run` gate), so with merge skew a release commit can publish 25 packages from a tree whose Rust tests never ran; a bad version cannot be unpublished. Direction: add `cargo test --workspace --locked` (with rust-cache) to `validate-release`, or gate publish on CI success for the same SHA.

## 3. Package unit tests never run on Windows or macOS

The macOS/Windows legs run only the CLI launcher tests and a native-binding load smoke (`ci.yml:99-109`; `pnpm test` is gated on `matrix.full`). Path-sensitive TS code (config root resolution, catalog path joining, next-plugin loaders) has no Windows unit-test coverage anywhere. Direction: run at least `config`, `core-node`, `next-plugin`, `vite-plugin` suites on the Windows leg — or record the trade-off deliberately.

swernerx:
## Summary

- Runs the existing pure-JS example/container contract suite as part of the repository test gate, so `scripts/container/start-plan.test.mjs` executes on the full Linux CI legs and release validation without requiring Docker.
- Runs focused config, core-node, next-plugin, and vite-plugin unit suites on the macOS and Windows validation legs after the native build. Browser/end-to-end and Linux-specific native packaging coverage remain in their dedicated gates.
- Makes release validation explicitly check out the triggering SHA, cache Rust build artifacts, and run `cargo test --workspace --locked` before the native publishing matrix can begin. Native and JavaScript publish jobs also explicitly check out that same SHA.
- Adds workflow-contract tests to prevent the test wiring, platform coverage, exact-SHA release checkouts, Rust cache, Rust test, and publish dependency from silently disappearing.
- Keeps the core-node test-support addon outside removable test fixtures, because Windows cannot unlink a loaded `.node` addon during Vitest cleanup, while reusing one stable ignored filename so repeated test runs cannot accumulate PID-suffixed native artifacts. The full core-node suite remains in the Windows gate.
- Uses Vitest's built-in test discovery for the Next and Vite plugin suites rather than a shell-expanded `src/*.test.ts` argument, so Windows runs all 93 plugin tests instead of treating the glob as a literal filter.
- Normalizes Vite's mocked catalog path fixture through the host resolver, so its native-boundary assertions cover both POSIX and Windows catalog paths without changing product behavior or reducing test coverage.

## Validation

- `RUSTUP_TOOLCHAIN=stable pnpm build`
- `RUSTUP_TOOLCHAIN=stable pnpm test`
- `CI=true RUSTUP_TOOLCHAIN=stable pnpm test:platform` (147 tests after rebasing onto current `main`)
- `pnpm check:workflow-contracts`
- `pnpm format:check`
- `pnpm lint`
- `RUSTUP_TOOLCHAIN=stable pnpm check-types`
- `RUSTUP_TOOLCHAIN=stable cargo test --workspace --locked`
- `pnpm check:release-set`
- YAML parsing for `.github/workflows/ci.yml` and `.github/workflows/publish.yml`

Closes #585


packages/core-node/src/index.test.ts
```diff
@@ -746,7 +746,15 @@ async function loadTestSupportAddon(tempDir: string): Promise<TestSupportBinding
...
   const libraryName = `${process.platform === "win32" ? "" : "lib"}palamedes_node.${extension}`
-  const addonPath = path.join(tempDir, "palamedes-node.node")
+  // Windows keeps loaded native addons open until Node exits, so a copy inside
+  // the per-test fixture would make its afterEach cleanup fail with EPERM.
+  // target/ is ignored and is already the build-artifact home for this addon.
+  const addonPath = path.join(
+    repoRoot,
+    "target",
+    "debug",
+    `palamedes-node-test-support-${process.pid}.node`
```
swernerx: This per-process filename leaves an ignored native addon in `target/debug` after every test process. In the fresh worktree, one platform run left a 28 MB `palamedes-node-test-support-5027.node`; because each new Node process gets a new PID and Windows cannot delete the loaded copy, repeated `pnpm test` or `pnpm test:platform` runs grow `target/` without bound. Please reuse a stable ignored filename once the previous process has exited, or use another cleanup design that keeps the strict `afterEach` fixture deletion intact.

swernerx (commented):
Verdict: changes requested

One actionable P2 finding is inline. The CI, release-gate, platform-discovery, path-normalization, and Remix-canary acceptance criteria otherwise check out at the exact head.

packages/core-node/src/index.test.ts
swernerx: Fixed in c2162640b7f786949973bbbed0892d941a02cba7.

The loaded addon now uses the single ignored `target/debug/palamedes-node-test-support.node` path, reused after the prior Node process exits; fixture directories remain subject to the existing strict `afterEach` cleanup. Added a regression test that repeated test-support setup resolves to that one stable filename, preventing PID-suffixed accumulation.

Validated: `RUSTUP_TOOLCHAIN=stable pnpm --filter @palamedes/core-node test` (21 tests), `RUSTUP_TOOLCHAIN=stable pnpm test:platform` (140 tests), `RUSTUP_TOOLCHAIN=stable pnpm test`, example/workflow contracts, format, lint, and type checks.

swernerx (commented):
Verdict: clear

The follow-up is scoped to the original finding. The fixed commit changes only `packages/core-node/src/index.test.ts`: the test-support loader now reuses the single ignored `target/debug/palamedes-node-test-support.node` path after the prior Node process exits, while the strict per-test fixture `afterEach` cleanup remains unchanged.

Independent validation at `c2162640b7f786949973bbbed0892d941a02cba7`:
- Two separate core-test processes passed 21/21; after each and after the broader runs, exactly one fixed-name addon existed with no PID/timestamp siblings.
- The regression assertion requires the exact fixed basename, and the loader obtains its sole addon path through that helper, so PID-unique naming cannot satisfy it.
- The platform contract passed 140/140 locally (20 config, 21 core-node, 53 Next, 46 Vite); the broader workspace suite, 14 example contracts, and 3 workflow contracts also passed.
- The exact-head Windows check passed those same four suites, including the loaded native-addon lifecycle, and all seven exact-head checks are successful.

No actionable P0-P2 findings.

packages/core-node/src/index.test.ts
swernerx: Rebased onto current `main`; the review fix is now at 199e3d515608e53db69cdefa85ce432e61ad0c93.

The semantic resolution retains the stable ignored `target/debug/palamedes-node-test-support.node` path, strict fixture cleanup, and the regression assertion preventing PID-suffixed accumulation, alongside main's current native-boundary tests.

Validated after rebuilding the local Darwin native package: core-node (28 tests plus packed-wrapper contract), platform gate (147 tests), example/workflow contracts, format, lint, and type checks.

PR #613, opened 2026-08-07
fix(cli): restrict npm tarball runtime files

### LINKED ISSUE #584: fix(cli): npm tarball ships build and test scripts
## Problem

`packages/cli/package.json` `files` includes `"scripts/"` wholesale. The launcher needs `scripts/run.mjs`, `scripts/platform.mjs`, `scripts/native.mjs` (imported by `bin/pmds`), but the directory also contains `build.mjs`, `build-native.mjs`, `platform.test.mjs`, `native.test.mjs`, `smoke.mjs` — build- and test-only files that ship in every install of the primary CLI package. All other published packages have tight `files` lists.

## Direction

Enumerate the three runtime files in `files` instead of the directory; verify with `pnpm pack` that the launcher still resolves.

swernerx:
Closes #584

## Summary

- Restrict `@palamedes/cli` to the `pmds` launcher and its three transitive runtime modules: `run.mjs`, `platform.mjs`, and `native.mjs`.
- Make the existing clean-tarball smoke test pack with `pnpm`, matching the publishing tool.
- Assert both the exact manifest allowlist and packed launcher runtime surface so package-list drift fails CI.

## Packed-file delta

`pnpm pack` now ships 7 files: `LICENSE`, `README.md`, `package.json`, `bin/pmds`, and the three required runtime modules. It removes the five non-runtime scripts: `build.mjs`, `build-native.mjs`, `platform.test.mjs`, `native.test.mjs`, and `smoke.mjs`.

## Validation

- `pnpm --filter @palamedes/cli build`
- `pnpm --filter @palamedes/cli test` (including clean extracted tarball, packed launcher version/flag/plugin paths, and native package)
- `pnpm pack --pack-destination …` plus `tar -tzf` inspection
- `pnpm --filter @palamedes/cli exec tsc --noEmit -p tsconfig.json`
- targeted `oxfmt` and ESLint
- `pnpm check:release-set`

### DIFF
--- a/packages/cli/scripts/smoke.mjs
+++ b/packages/cli/scripts/smoke.mjs
@@ -233,7 +233,7 @@ function assertPackedCliRuntimeFiles(archivePath) {
 
   const packedFiles = execFileSync("tar", ["-tzf", archivePath], { encoding: "utf8" })
     .trim()
-    .split("\n")
+    .split(/\r?\n/u)
     .filter(Boolean)
     .sort()
   const expectedRuntimeFiles = [
--- a/packages/cli/scripts/smoke.mjs
+++ b/packages/cli/scripts/smoke.mjs
@@ -17,6 +17,23 @@ import { resolvePlatformPackage } from "./platform.mjs"
 
 const packageDir = path.resolve(import.meta.dirname, "..")
 const repoRoot = path.resolve(packageDir, "../..")
+const expectedCliManifestFiles = [
+  "LICENSE",
+  "README.md",
+  "bin/pmds",
+  "scripts/run.mjs",
+  "scripts/platform.mjs",
+  "scripts/native.mjs",
+]
+const expectedPackedCliFiles = [
+  "package/LICENSE",
+  "package/README.md",
+  "package/bin/pmds",
+  "package/package.json",
+  "package/scripts/native.mjs",
+  "package/scripts/platform.mjs",
+  "package/scripts/run.mjs",
+]
 let platformPackage
 try {
   platformPackage = resolvePlatformPackage()
@@ -52,8 +69,8 @@ try {
   // node_modules tree. No lifecycle hook runs while constructing this fixture.
   const archiveDir = path.join(fixtureRoot, "archives")
   mkdirSync(archiveDir)
-  const cliArchive = packPackage(packageDir, archiveDir)
-  const nativeArchive = packPackage(platformPackageDir, archiveDir)
+  const cliArchive = packPackage(packageDir, archiveDir, "pnpm")
+  const nativeArchive = packPackage(platformPackageDir, archiveDir, "npm")
   assertPackedCliRuntimeFiles(cliArchive)
   const cliInstallDir = path.join(fixtureRoot, "node_modules", "@palamedes", "cli")
   const nativeInstallDir = path.join(fixtureRoot, "node_modules", ...platformPackage.split("/"))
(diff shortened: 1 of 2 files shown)


packages/cli/scripts/smoke.mjs
```diff
@@ -189,7 +190,7 @@ if (output !== nativeOutput) {
 
 function packPackage(packagePath, archiveDir) {
   const before = new Set(readdirSync(archiveDir))
-  const packageManager = process.platform === "win32" ? "npm.cmd" : "npm"
+  const packageManager = process.platform === "win32" ? "pnpm.cmd" : "pnpm"
```
swernerx: The smoke now uses pnpm pack for both archives, but the release script publishes native artifact packages with npm publish. On Darwin the same built native package is 4 files / 8,792,354 bytes with pnpm (it adds LICENSE), versus the real npm artifact at 3 files / 8,708,557 bytes. Pass the package manager into packPackage so the CLI uses pnpm and the native package uses npm; otherwise this can pass against an artifact the release never publishes.

packages/cli/scripts/smoke.mjs
```diff
@@ -211,3 +212,44 @@ function extractPackage(archivePath, destination) {
...
+    .split(/\r?\n/u)
+    .filter(Boolean)
+    .sort()
+  const expectedRuntimeFiles = [
+    "package/bin/pmds",
+    "package/scripts/native.mjs",
+    "package/scripts/platform.mjs",
+    "package/scripts/run.mjs",
+  ]
+  const packedRuntimeFiles = packedFiles.filter(
```
swernerx: This filters away every non-runtime entry before comparison, so listings with package.json, README.md, and LICENSE all missing, or with unexpected top-level files, still pass. The preceding trim also accepts a first or last path whose whitespace was corrupted. My negative entry mutations reproduced all of those passes. Keep the CRLF split, remove only the single terminal empty record produced by a valid listing, and compare the full sorted seven-entry tar list exactly.

swernerx (commented):
Verdict: changes requested

The CLI allowlist itself is correct and the pnpm CLI tarball resolves the complete launcher import graph, but two packed-artifact checks still diverge from the release contract. Details are inline.

packages/cli/scripts/smoke.mjs
swernerx: Fixed in a05bb178dc6ffcddb7baccc96d7866ddd3aee518. `packPackage` now receives the publisher explicitly: the CLI uses `pnpm pack`, while the native package uses `npm pack`, matching `publish-package-if-needed.mjs`. The clean fixture still extracts both artifacts and exercises native bin, launcher `version`, `--version`, and plugin paths. On Darwin, the verified native npm tarball contains exactly `README.md`, `package.json`, and `bin/pmds` (3 entries, 8,769,246 bytes).

packages/cli/scripts/smoke.mjs
swernerx: Fixed in a05bb178dc6ffcddb7baccc96d7866ddd3aee518. The assertion now parses CRLF-safe tar output, removes only the one terminal empty record, and compares the complete sorted CLI tar listing exactly: `LICENSE`, `README.md`, `package.json`, `bin/pmds`, and the three launcher modules. It includes negative mutations for removing every expected entry, adding an unexpected top-level entry, and leading/trailing whitespace corruption. The verified pnpm CLI tarball has exactly those 7 entries.

swernerx (commented):
Verdict: clear

Both original findings are addressed at the exact reviewed head. The smoke now uses the same package manager as each publisher path, asserts the complete seven-entry CLI artifact with strict CRLF-only line-ending handling, and exercises the extracted launcher, native, and plugin routes without workspace leakage.

PR #615, opened 2026-08-07
fix(core-node): harden native boundary contracts

### LINKED ISSUE #582: fix(core-node): boundary contract polish — exact platform pinning + version assert, dead include/exclude fields, lone-surrogate guard, sync-only decision
Four boundary-contract items, cheap individually, coherent as one pass:

## 1. Platform binary packages are ^-range coupled with no runtime cross-check

`packages/core-node/package.json:50-55` declares the five platform packages as `workspace:^`, which `pnpm publish` rewrites to `^1.14.0`. Resolution can legally pair wrapper `1.14.0` with binary `1.20.x` (fresh install after later releases, partially-updated lockfile), and `loadNativeBindings` never compares `getNativeInfo().palamedesVersion` against the wrapper version — a skewed pair fails at call time with `native.applyTranslationPatches is not a function` instead of a clear versioned load error. napi convention is exact pinning. Direction: `workspace:*` (publishes exact) plus a version assert at load. Same applies to `packages/cli`'s platform packages if ranged.

## 2. The native contract requires `include`/accepts `exclude` but discards both

`crates/palamedes-node/src/catalog_config.rs:46-58` — `let _ = value.include; let _ = value.exclude;`; the generated type makes `include` required. Only the Vite plugin applies these filters TS-side. Callers of `auditCatalogs`/`listTranslationCandidates`/`compileCatalogArtifact` reasonably believe it scopes the native operation. Direction: document as host-side-only via doc comments on the generated structs, or make them optional.

## 3. Lone surrogates are silently rewritten to U+FFFD at the string boundary

All paths cross as `String` (V8 replaces unpaired surrogates), so a path/message containing a lone surrogate becomes a *different* string in Rust — manifesting as `Catalog file not found` / `translation.unknown_message` for inputs that print identically. Direction: cheap guard in the TS wrapper rejecting request strings that fail `String.prototype.isWellFormed()`.

## 4. The entire native surface is synchronous — record the decision

No async/AsyncTask anywhere in the binding; multi-file workflows (extract, audit, artifact compile, patch apply) block the Node event loop, and the Vite plugin calls `compileCatalogModule` inside module hooks on the dev-server thread. ADR-009 is silent on sync-vs-async, so this is an unstated tradeoff. Direction: either add Async-suffixed variants for the heavy workflows or note the sync-boundary decision in ADR-009 explicitly.

swernerx:
Closes #582

## Boundary contract changes

- Uses `workspace:*` for core-node and CLI platform optional dependencies, which pnpm publishes as exact same-release pins. The core-node loader now verifies `getNativeInfo().palamedesVersion` before exposing the binding and gives a versioned reinstall/remediation error on skew.
- Makes generated catalog `include` optional and documents both `include` and `exclude` as host-side-only fields. Vite keeps applying its existing source filter before native compilation.
- Guards every call through the Node binding with a linear, non-mutating well-formed-Unicode traversal. Lone UTF-16 surrogates fail with the operation and field path; valid pairs, combining text, nested arrays, records, and Maps remain accepted.
- Records the synchronous N-API decision in ADR-009: calls can block the Node thread, Vite uses them in bounded host hooks, and async variants need measured event-loop pressure plus an explicit lifecycle contract.

## Validation

- `pnpm lint` and `pnpm format:check`
- Core-node Unicode/version-fixture/packed-manifest tests; CLI platform and packed-manifest smoke tests; Vite host-filtering tests (46 passing)
- `pnpm check:release-set` and generated native type check
- `RUSTUP_TOOLCHAIN=1.95 cargo fmt --all --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, and `cargo test --workspace --locked`

### DIFF
--- a/packages/core-node/src/native-loader.ts
+++ b/packages/core-node/src/native-loader.ts
@@ -190,7 +190,8 @@ export function snapshotNativeArguments(operation: string, arguments_: unknown[]
       continue
     }
 
-    if (value instanceof Map) {
+    const classification = classifySnapshotObject(value, currentPath)
+    if (classification.kind === "map") {
       const mapSnapshot = new Map<unknown, unknown>()
       snapshots.set(value, mapSnapshot)
       current.assign(mapSnapshot)
@@ -227,13 +228,12 @@ export function snapshotNativeArguments(operation: string, arguments_: unknown[]
       }
       continue
     }
-    if (Array.isArray(value)) {
+    if (classification.kind === "array") {
       const arraySnapshot: unknown[] = []
-      arraySnapshot.length = value.length
+      arraySnapshot.length = classification.length
       snapshots.set(value, arraySnapshot)
       current.assign(arraySnapshot)
-      for (const key of ownPropertyNames(value, currentPath).reverse()) {
-        if (key === "length") continue
+      for (const key of enumerablePropertyNames(value, currentPath).reverse()) {
         const propertyPath = appendNativeArgumentPath(currentPath, arrayPropertyPath(key))
         assertWellFormedPropertyName(key, propertyPath)
         pending.push({
@@ -249,16 +249,10 @@ export function snapshotNativeArguments(operation: string, arguments_: unknown[]
       continue
     }
 
-    const prototype = objectPrototype(value, currentPath)
-    if (prototype !== Object.prototype && prototype !== null) {
-      throw new TypeError(
-        `Palamedes native boundary requires a plain object at ${formatNativeArgumentPath(currentPath)}; copy class instances into request data before calling ${operation}.`
-      )
-    }
-    const snapshot = Object.create(prototype) as Record<string, unknown>
-    snapshots.set(value, snapshot)
-    current.assign(snapshot)
-    for (const key of ownPropertyNames(value, currentPath).reverse()) {
+    const recordSnapshot: Record<string, unknown> = {}
+    snapshots.set(value, recordSnapshot)
+    current.assign(recordSnapshot)
+    for (const key of enumerablePropertyNames(value, currentPath).reverse()) {
       const propertyPath = appendNativeArgumentPath(currentPath, `.${key}`)
       assertWellFormedPropertyName(key, propertyPath)
       pending.push({
@@ -267,7 +261,7 @@ export function snapshotNativeArguments(operation: string, arguments_: unknown[]
         key,
         path: propertyPath,
         assign(snapshotValue: unknown) {
-          defineSnapshotProperty(snapshot, key, snapshotValue)
+          defineSnapshotProperty(recordSnapshot, key, snapshotValue)
         },
       })
     }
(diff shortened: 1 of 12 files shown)


packages/core-node/src/native-loader.ts
```diff
@@ -0,0 +1,251 @@
...
+    if (Array.isArray(value)) {
+      for (let index = value.length - 1; index >= 0; index -= 1) {
+        pending.push({
+          value: value[index],
+          path: appendNativeArgumentPath(currentPath, `[${index}]`),
+        })
+      }
+      continue
+    }
+    const objectEntries = Object.entries(value)
```
swernerx: The guard validates accessor-backed properties on one read, but N-API converts the original object and reads them again. A getter that returns a valid string first and U+D800 second makes `renderCatalogModule()` emit U+FFFD instead of throwing, so the silent-rewrite failure remains reachable through the public API. Please validate and pass a single plain snapshot of the arguments (or otherwise guarantee the native call consumes the exact values checked), and cover an accessor/proxy that changes between reads.

swernerx (commented):
Verdict: changes requested

The exact platform pinning, load-time version check, generated include/exclude contract, Vite host filtering, and sync-only ADR line up with the implementation and packed artifacts. One P2 Unicode-boundary bypass remains in the preflight; details are inline.

packages/core-node/src/native-loader.ts
swernerx: Fixed in 1b60ed108d5e3c2c044c4aaca8781ea74874d3df.

The guarded binding now snapshots and Unicode-validates the exact plain arguments supplied to N-API, so accessor and Proxy values are read once and cannot change after validation. Added the real renderCatalogModule accessor/Proxy reproduction plus nested arrays, Maps, cycles, lone high/low surrogates, valid pairs/combining text, stable field paths, getter-failure cause, and no-mutation coverage. Focused core-node + packed artifact, CLI packed consumer/platform, Vite, generated types, release-set, lint/format, and Rust 1.95 workspace clippy/tests pass.

packages/core-node/src/native-loader.ts
```diff
@@ -0,0 +1,402 @@
...
+      continue
+    }
+
+    const existingSnapshot = snapshots.get(value)
+    if (existingSnapshot !== undefined) {
+      current.assign(existingSnapshot)
+      continue
+    }
+
+    if (value instanceof Map) {
```
swernerx: `value instanceof Map` invokes a Proxy's `getPrototypeOf` trap before any error wrapper, and the normal-object path invokes the same trap again in `objectPrototype`. Through public `renderCatalogModule`, a plain-object Proxy was probed twice; a trap that throws on the first probe escaped as the original `Error` with no boundary path or `cause`. The direct array `length` read has the same unwrapped failure mode. Please classify objects and read array metadata through single guarded operations, then cover one-call proxy traps and stable boundary errors for prototype and length failures.

packages/core-node/src/native-loader.ts
```diff
@@ -0,0 +1,402 @@
...
+          assign(snapshotValue: unknown) {
+            defineSnapshotProperty(arraySnapshot, key, snapshotValue)
+          },
+        })
+      }
+      continue
+    }
+
+    const prototype = objectPrototype(value, currentPath)
+    if (prototype !== Object.prototype && prototype !== null) {
```
swernerx: The snapshot changes native-visible object semantics: the prototype gate rejects a class-backed record that the native `renderCatalogModule` binding accepts, while `Object.getOwnPropertyNames` plus `enumerable: true` promotes a non-enumerable message key that the raw binding ignores and causes the public wrapper to emit it. Please copy the supported native-visible own enumerable record values into a plain snapshot without rejecting solely because the source has a custom prototype, and add public regressions for both cases.

swernerx (commented):
Verdict: changes requested

The exact `renderCatalogModule` accessor reproduction is fixed: the getter is read once, N-API receives the first validated value, and no U+FFFD is emitted. The focused 24-test suite and broader repository gates pass. Two P2 snapshot-contract regressions remain inline. GitHub also currently reports this head as DIRTY/CONFLICTING against the advanced `main`, so the required mergeability gate is not satisfied.

packages/core-node/src/native-loader.ts
swernerx: Fixed in b86d51b2bc6acad5e4ec0ae5902bf1c8c6198864.

Classification now performs one guarded source-prototype read (rather than instanceof plus a second probe), and array length is read once through the same deterministic boundary error translation. Added public renderCatalogModule coverage for a one-call getPrototypeOf trap and public prototype/array-length trap failures that assert the stable field path and original cause. The focused 26-test core suite and packed artifact test pass.

packages/core-node/src/native-loader.ts
swernerx: Fixed in b86d51b2bc6acad5e4ec0ae5902bf1c8c6198864.

Snapshots now copy only native-visible own enumerable record values into a fresh plain object, without rejecting a class-backed record solely for its prototype or promoting non-enumerable properties. Added public renderCatalogModule regressions for both class-backed and non-enumerable records. Generated-type checks, lint/format, release-set, CLI packed consumer/platform, Vite, and Rust 1.95 workspace gates pass.

packages/core-node/src/native-loader.ts
swernerx: Rebased onto current main; the final fix SHA is 9f85d8af400e28dafa39a89b8f84c21cbb9cc2ff (the earlier b86d51b SHA was rewritten by the required rebase). The guarded one-prototype/one-length reads and public Proxy error regressions remain unchanged; rebased core packed, CLI packed consumer/platform, Vite, generated-type, lint/format, release, and Rust 1.95 gates pass.

packages/core-node/src/native-loader.ts
swernerx: Rebased onto current main; the final fix SHA is 9f85d8af400e28dafa39a89b8f84c21cbb9cc2ff (the earlier b86d51b SHA was rewritten by the required rebase). The enumerable plain-record snapshot semantics and public class/non-enumerable regressions remain unchanged; rebased core packed, CLI packed consumer/platform, Vite, generated-type, lint/format, release, and Rust 1.95 gates pass.

swernerx (commented):
Verdict: clear

All three prior findings are resolved at the rebased head. The native binding consumes the one validated snapshot; Proxy metadata and property failures are read once and translated with stable paths and causes; class-backed records remain supported; and non-enumerable keys stay native-invisible. The rebase also retains both the current-main CLI tarball-surface smoke and the exact five-platform version assertion. Focused, adversarial, packed-consumer, TypeScript, Vite, release, lint/format, LLMS, and Rust workspace gates pass with no remaining P0-P2 finding.

PR #616, opened 2026-08-07
fix(core): keep translation workflows target-only

### LINKED ISSUE #581: fix(core): translation workflow consistency — source locale is not filtered from explicit requests; one missing catalog file fails the whole enumeration
## 1. Source locale is not excluded from explicit candidate listing or patching

`selected_locales` (`crates/palamedes/src/translation_candidates.rs:595-613`) filters `source_locale` only in the default (empty-request) branch; the patch path (`:414-425`) accepts any configured locale. Reproduced: `locales: ["en"]` with `sourceLocale: "en"` returns en→en candidates (every empty-`msgstr` source entry reported as "missing translation"), and a patch batch can silently rewrite source-catalog `msgstr` values that the next `extract` then fights with. Contrast `catalog_audit.rs:365-368`, which filters the source locale from explicit requests too.

Direction: filter `locale != source_locale` in both branches and reject source-locale patch ids with a structured diagnostic.

## 2. `list_translation_candidates` hard-fails when any configured locale catalog file is missing

`:621-624` propagates `ReadFile` for a missing file while iterating every configured non-source locale. Adding a locale to the config before its first `pmds extract` (fresh branch/clone) makes the whole workflow API error out. Contrast `catalog_audit.rs:387-389`, which skips missing files and reports `catalog.missing_locale`.

Direction: skip missing catalogs with a structured `translation.missing_catalog` diagnostic in the default flow; keep hard errors for explicitly targeted locales.

*Filed from the 2026-08-07 repository audit (both reproduced by execution).*

swernerx:
Closes #581

## What changed

- Excludes the configured source locale from default and explicit candidate locale selection. Source candidate IDs and patch IDs now return the structured `translation.source_locale` diagnostic before any catalog write.
- Default candidate enumeration now skips only `NotFound` target catalogs and emits deterministic `translation.missing_catalog` diagnostics with actionable `locale` and `catalogPath` context. Explicitly selected missing locales still return the existing hard `ReadFile` error; permission, parse, and other I/O failures remain hard errors.
- Adds the additive diagnostic fields to the NAPI and `@palamedes/core-node` public contract and documents the behavior.

## Validation

- `cargo +1.95 test -p palamedes translation_candidates`
- `RUSTUP_TOOLCHAIN=1.95 pnpm --filter @palamedes/core-node generate-native-types`
- `RUSTUP_TOOLCHAIN=1.95 pnpm --filter @palamedes/core-node test`
- `cargo +1.95 fmt --all --check`
- `cargo +1.95 clippy --workspace --all-targets --locked -- -D warnings`
- `cargo +1.95 test --workspace --locked -q`
- `cargo +stable fmt --all --check`
- `cargo +stable clippy --workspace --all-targets --locked -- -D warnings`
- `cargo +stable test --workspace --locked -q`
- `RUSTUP_TOOLCHAIN=1.95 pnpm --filter @palamedes/core-node check-native-types`

### DIFF
--- a/crates/palamedes/src/translation_candidates.rs
+++ b/crates/palamedes/src/translation_candidates.rs
@@ -251,6 +251,12 @@ pub struct TranslationWorkflowDiagnostic {
     /// Associated candidate identity, when available.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub id: Option<TranslationCandidateId>,
+    /// Resolved catalog path associated with this diagnostic, when applicable.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub catalog_path: Option<String>,
+    /// Target locale associated with this diagnostic, when applicable.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub locale: Option<String>,
 }
 
 #[derive(Debug)]
@@ -290,10 +296,15 @@ pub fn list_translation_candidates(
     request: &TranslationCandidateRequest,
 ) -> PalamedesResult<TranslationCandidateResult> {
     let locales = selected_locales(&request.config, &request.locales);
+    let default_locale_selection = request.locales.is_empty();
     let explicit = !request.targets.is_empty();
     let mut requested = BTreeMap::<TranslationCandidateId, usize>::new();
     let mut diagnostics = Vec::new();
     for target in &request.targets {
+        if is_source_locale(&request.config, &target.locale) {
+            diagnostics.push(source_locale_diagnostic(target.clone()));
+            continue;
+        }
         let count = requested.entry(target.clone()).or_default();
         *count += 1;
         if *count == 2 {
@@ -309,7 +320,17 @@ pub fn list_translation_candidates(
     let mut seen = BTreeSet::new();
     for catalog in &request.config.catalogs {
         for locale in &locales {
-            let loaded = load_catalog(&request.config, catalog, locale)?;
+            let loaded = match load_catalog(&request.config, catalog, locale) {
+                Ok(loaded) => loaded,
+                Err(PalamedesError::ReadFile { path, source })
+                    if default_locale_selection
+                        && source.kind() == std::io::ErrorKind::NotFound =>
+                {
+                    diagnostics.push(missing_catalog_diagnostic(locale, path));
+                    continue;
+                }
+                Err(error) => return Err(error),
+            };
             for (key, message) in &loaded.messages {
                 let id = TranslationCandidateId {
                     catalog: loaded.scope.clone(),
@@ -400,6 +421,12 @@ where
     let mut groups = BTreeMap::<CatalogBatchKey, Vec<usize>>::new();
 
     for (index, patch) in request.patches.iter().enumerate() {
+        if is_source_locale(&request.config, &patch.id.locale) {
+            rejected.insert(index);
+            diagnostics.push(source_locale_diagnostic(patch.id.clone()));
+            continue;
+        }
+
         for message in validate_plural_translation_branches(&patch.translation) {
             rejected.insert(index);
             diagnostics.push(workflow_diagnostic(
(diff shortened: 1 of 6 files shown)


crates/palamedes/src/translation_candidates.rs
```diff
@@ -290,10 +296,15 @@ pub fn list_translation_candidates(
     request: &TranslationCandidateRequest,
 ) -> PalamedesResult<TranslationCandidateResult> {
     let locales = selected_locales(&request.config, &request.locales);
+    let default_locale_selection = request.locales.is_empty();
```
swernerx: An exact `targets` request still takes the default-selection path when `locales` is empty. If that target's catalog is missing, this branch swallows `ReadFile(NotFound)` and returns `translation.missing_catalog` plus `translation.unknown_target`; the docs promise an explicitly requested target remains a hard read error. Only skip NotFound when both `locales` and `targets` are empty, and add the `targets: [missingId]` regression case.

crates/palamedes/src/translation_candidates.rs
```diff
@@ -400,6 +421,12 @@ where
     let mut groups = BTreeMap::<CatalogBatchKey, Vec<usize>>::new();
 
     for (index, patch) in request.patches.iter().enumerate() {
+        if is_source_locale(&request.config, &patch.id.locale) {
```
swernerx, line 424: Recording this rejection before grouping the source patch is not enough to make it an early batch result. The function still loads every other group before the `!rejected.is_empty()` return, so a mixed source-locale patch plus an unreadable target catalog throws `ReadFile` and loses the required structured `translation.source_locale` result. Return the preliminary validation result before entering the catalog-load loop, marking the other patches `notApplied`, and cover the mixed unreadable-target case.

swernerx (commented):
Verdict: changes requested

The exact-string source-locale exclusion, NotFound-only catch, deterministic diagnostics, and N-API/Node field propagation are in good shape. Two explicit-scope and early-validation gaps remain; both are on the inline threads. The existing Rust/Node suites and both Rust 1.95/stable matrices pass, but adversarial exact-head probes reproduce both behaviors.

crates/palamedes/src/translation_candidates.rs
swernerx: Fixed in c70f334d222b6533362d548caf12a5d9388ff0f9.

Default missing-catalog skipping now requires both `locales` and `targets` to be empty. The regression covers `locales: []` with `targets: [missingId]` and verifies the original `ReadFile(NotFound)` hard error is preserved.

Focused evidence: `cargo +1.95 test -p palamedes translation_candidates` (14 passed); full Rust 1.95/stable workspace format, Clippy, and tests also passed.

crates/palamedes/src/translation_candidates.rs
swernerx: Fixed in c70f334d222b6533362d548caf12a5d9388ff0f9.

Preliminary patch validation rejections now return before the catalog-load loop, with all other patches reported as `notApplied`. The new regression combines a tampered source-locale patch with a directory-backed unreadable target catalog and verifies the structured `translation.source_locale` result is retained.

Focused evidence: `cargo +1.95 test -p palamedes translation_candidates` (14 passed); `@palamedes/core-node` (21 tests), generated-type drift, package TypeScript, and full Rust 1.95/stable workspace checks also passed.

swernerx (commented):
Verdict: clear

Both prior findings are addressed at c70f334d222b6533362d548caf12a5d9388ff0f9. Explicit missing targets now keep the hard ReadFile/NotFound failure, and preliminary source-locale rejections return before any catalog I/O, with the remaining patches deterministically marked notApplied and no writes. Default enumeration, valid-target atomicity and ordering, missing-catalog diagnostics, and the Node/N-API/generated type boundary remain intact. The focused adversarial tests and the Rust 1.95/stable and Node validation matrices are green.

PR #621, opened 2026-08-07
fix(core): preserve live macro imports

### LINKED ISSUE #599: Make macro import erasure usage-aware
## Problem

Palamedes currently removes an entire macro import declaration whenever the file contains at least one macro replacement. Import cleanup is therefore tied to whether the transform changed the file, not to whether each imported binding still has surviving references.

```tsx
import { t } from "@palamedes/core/macro";

export function Example() {
  const label = t`Upload failed`;
  console.log(t);
  return label;
}
```

The tagged template is correctly rewritten, but the complete macro import is removed and `console.log(t)` is left behind. The emitted module reads an unbound identifier and fails with `ReferenceError: t is not defined`.

The same `console.log(t)` source remains bound when the file contains no transformed macro use because the transform then leaves the import untouched. Output correctness should not depend on an unrelated use of the same binding elsewhere in the file.

The macro entry points are deliberately small runtime stubs and declare `sideEffects: false`. Keeping an import when a reference survives is therefore a safe fallback: a bundler can remove fully unused imports, while a genuinely executed untransformed macro produces the package's intentional compiler-configuration error instead of an accidental unbound-identifier error.

## Desired behavior

Macro import cleanup should be binding- and usage-aware:

1. Recognized macro syntax is transformed as it is today.
2. An imported macro specifier is removed only when every reference to that binding was consumed by the transform or deliberately rebound.
3. If any reference survives in emitted JavaScript, its import specifier remains so the transform never introduces an unbound identifier.
4. When precise reference information is unavailable, preserve the import conservatively.

This must work per specifier rather than per import declaration. It must also cover aliases, multiple macro imports in one declaration, shadowed locals, and type-only references.

`Trans` needs its existing special handling: transformed uses are rebound to the compiled framework component. Any surviving references must remain bound without creating a duplicate local import.

## Architecture

Implement usage-aware cleanup from Palamedes' shared native OXC parse, binding, and transform facts. Do not introduce a second TypeScript/ESLint parser or a repository-specific compatibility layer.

## Acceptance criteria

- [ ] Transforming one macro use never unbinds another surviving reference to the same imported symbol.
- [ ] Import cleanup is decided per binding/specifier, not by deleting every macro import declaration after the first replacement.
- [ ] Fully consumed macro bindings can still be removed from emitted source; preserving them conservatively must remain safe for tree-shaking because the macro packages are side-effect-free.
- [ ] Focused fixtures cover runtime-only use, mixed transformed/runtime use, hook dependency arrays, object and argument positions, aliases, multiple specifiers, shadowed locals, type-only references, and `Trans` rebinding.
- [ ] The generated module contains no duplicate bindings and no transform-created unbound macro identifiers.
- [ ] The implementation shares Palamedes' native source facts; there is no second parser or file-discovery model.

## Non-goals

- Detecting arbitrary untranslated user-visible strings.
- Detecting or banning macro bindings used as ordinary runtime values.
- Reintroducing the retired Lingui ESLint plugin or branded technical-string types.
- Requiring a repository-specific lint rule.

## References

- Parent architecture and Core responsibility: #510

swernerx:
## Summary

Make macro cleanup binding-aware with the shared native OXC semantic facts. Transformed bindings are removed only after every value reference is consumed; live aliases/specifiers stay imported. `Trans` uses a collision-free compiled alias when its macro binding remains live.

## Issue

Closes #599

## Validation

- `cargo +1.95 fmt --all --check`
- `cargo +1.95 clippy --workspace --all-targets --locked -- -D warnings`
- `cargo +1.95 test --workspace --locked`
- `pnpm format:check` and `pnpm lint`
- Public native transform reproduction

## Follow-up

None.

### DIFF
--- a/crates/palamedes/src/transform/imports.rs
+++ b/crates/palamedes/src/transform/imports.rs
@@ -1,18 +1,22 @@
 use std::collections::{HashMap, HashSet};
 
-use crate::source_macros::record_macro_import_declaration;
 pub(super) use crate::source_macros::ImportedMacro;
+use crate::{source_macros::record_macro_import_declaration, transform::Replacement};
 use oxc_ast::ast::{
     BindingIdentifier, IdentifierReference, ImportDeclaration, ImportDeclarationSpecifier,
     ImportOrExportKind, JSXIdentifier,
 };
 use oxc_ast_visit::{walk, Visit};
+use oxc_semantic::Semantic;
+use oxc_span::GetSpan;
+use oxc_syntax::symbol::SymbolId;
 
 pub(super) struct ImportCollector {
     runtime_module: String,
     runtime_import_name: String,
     pub macro_imports: HashMap<String, ImportedMacro>,
-    pub macro_import_ranges: Vec<(usize, usize)>,
+    macro_specifiers: Vec<MacroImportSpecifier>,
+    reference_symbols: HashMap<(u32, u32), SymbolId>,
     pub removed_macro_import: Option<(String, usize)>,
     pub has_reusable_runtime_import: bool,
     /// Number of bindings that could shadow calls through the configured local name.
@@ -28,7 +32,8 @@ impl ImportCollector {
             runtime_module: runtime_module.to_string(),
             runtime_import_name: runtime_import_name.to_string(),
             macro_imports: HashMap::new(),
-            macro_import_ranges: Vec::new(),
+            macro_specifiers: Vec::new(),
+            reference_symbols: HashMap::new(),
             removed_macro_import: None,
             has_reusable_runtime_import: false,
             runtime_import_binding_count: 0,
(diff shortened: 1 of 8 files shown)


swernerx (commented):
Verdict: changes requested

The binding-aware OXC direction is right, and the exact mixed-use `t` repro now works end to end: the transformed template executes, the surviving `t` stays bound, and calling it reaches the package's intentional runtime-stub error instead of `ReferenceError`. The full Rust and JS/TS validation matrix is green.

Three compiler-proven P1 cases still violate the issue contract: a same-binding type query is left unbound, an unrelated authored `Trans` can collide with the generated compiled import, and a type-only or differently imported compiled symbol can be mistaken for a reusable runtime `Trans`. I left the minimal repro and correction for each inline. The Windows-only path adjustment is test-only and passes on Windows CI, but the surrounding matcher is still partial rather than exact.

crates/palamedes/src/transform/imports.rs
```diff
@@ -28,14 +32,203 @@ impl ImportCollector {
...
+    ) -> Vec<Replacement> {
+        let mut removable = HashSet::new();
+        for (index, specifier) in self.macro_specifiers.iter().enumerate() {
+            let Some(symbol_id) = specifier.symbol_id else {
+                continue;
+            };
+            let all_value_references_consumed = semantic
+                .scoping()
+                .get_resolved_references(symbol_id)
+                .filter(|reference| reference.is_value())
```
swernerx: [P1] Preserve surviving type references to this binding

Filtering to value references makes the following `all(...)` ignore a live type-namespace use. With `import { t } from "@palamedes/core/macro"; type MacroValue = typeof t; function Example() { return t\`Hello\`; }`, the public transform deletes the macro import but leaves `typeof t`; TypeScript then reports TS2304. Treat every unconsumed resolved reference as live, or preserve the specifier conservatively, and add a same-binding type-query regression.

crates/palamedes/src/transform/visitor.rs
```diff
@@ -74,6 +80,35 @@ impl<'a> TransformVisitor<'a> {
...
+                semantic,
+                symbol_id,
+                &self.consumed_binding_ranges,
+            ) {
+                aliases
+                    .entry(symbol_id)
+                    .or_insert_with(|| unique_identifier("__palamedesTrans", &mut reserved))
+                    .clone()
+            } else {
+                "Trans".to_string()
```
swernerx, line 114: [P1] Avoid claiming an authored `Trans` binding

This branch always chooses local name `Trans` when the macro binding is fully consumed, even if an unrelated authored binding already owns that name. `import { Trans as MacroTrans } from "@palamedes/react/macro"; const Trans = () => null; <MacroTrans>Hello</MacroTrans>` emits both `import { Trans } from "@palamedes/react/compiled"` and `const Trans`, and TypeScript reports TS2440. Use a collision-free generated alias whenever `Trans` is reserved by anything other than the exact removable macro binding, with a regression for the aliased-macro/local-`Trans` case.

crates/palamedes/src/transform/mod.rs
```diff
@@ -290,15 +300,18 @@ pub fn transform_macros(
...
-        .cloned()
-        .collect::<Vec<_>>();
-    trans_modules.sort();
-    for module in trans_modules {
-        if !collector.trans_import_sources.contains(&module) {
-            let _ = writeln!(prefix, "import {{ Trans }} from \"{module}\";");
+    let mut trans_imports = visitor.trans_imports.iter().cloned().collect::<Vec<_>>();
+    trans_imports.sort();
+    for (module, local_name) in trans_imports {
+        if local_name != "Trans" || !collector.trans_import_sources.contains(&module) {
```
swernerx: [P1] Verify the reusable compiled `Trans` is a value import of that export

This skip trusts only the collected module/local-name pair. A pre-existing `import type { Trans } from "@palamedes/react/compiled"` suppresses the generated value import and leaves `<Trans ...>` with TS1361; `import { Fragment as Trans }` silently routes transformed output to the wrong component. Record reuse only for a value declaration/specifier whose imported name is exactly `Trans`, and cover both type-only and renamed imports.

packages/core-node/src/index.test.ts
```diff
@@ -563,7 +563,7 @@ msgstr ""
       expect.objectContaining({
         code: "translation.missing_catalog",
         locale: "fr",
-        catalogPath: path.join(rootDir, "locales", "fr", "messages.po"),
+        catalogPath: `${rootDir}${path.sep}locales/fr/messages.po`,
```
swernerx: [P2] Keep the Windows diagnostic assertion exact

The outer `toStrictEqual([...])` fixes one-element cardinality, but `expect.objectContaining(...)` still accepts extra fields and omits the diagnostic's `message`/`id` contract. Replace the partial matcher with the complete expected diagnostic object while retaining this portable path spelling, so the test proves all fields and remains strictly test-only.

crates/palamedes/src/transform/imports.rs
swernerx: Fixed in 569b73b0e70246b0e7f825482c158f2ed047d2a5. Macro cleanup now treats every resolved, unconsumed reference as live, including type-namespace references such as `typeof t`; the macro specifier is retained in that case. Added `preserves_macro_binding_for_same_binding_type_queries`. Evidence: focused transform suite 97/97 passed; full Rust 1.95/stable workspace gates and full JS/TS validation passed.

crates/palamedes/src/transform/visitor.rs
swernerx: Fixed in 569b73b0e70246b0e7f825482c158f2ed047d2a5. Generated compiled `Trans` now aliases collision-free whenever another authored binding owns `Trans`, while still using bare `Trans` when the exact macro binding is removable. Added `aliases_compiled_trans_when_an_authored_trans_binding_is_reserved` for aliased macro/local-`Trans`. Evidence: focused transform suite 97/97 passed; rebuilt public core-node replay produced `<__palamedesTrans>` alongside the authored local without a duplicate binding.

crates/palamedes/src/transform/mod.rs
swernerx: Fixed in 569b73b0e70246b0e7f825482c158f2ed047d2a5. Reuse is now recorded only for a value declaration/specifier whose source is the compiled framework module and whose imported and local names are both `Trans`. Type-only `Trans` and `Fragment as Trans` now force a collision-free generated value alias. Added `reuses_only_exact_value_trans_imports`, including both negative cases and the exact-value positive case. Evidence: focused transform suite 97/97 passed; rebuilt public core-node replays verified both negative imports emit/use `__palamedesTrans` and the exact value import is reused once.

packages/core-node/src/index.test.ts
swernerx: Fixed in 569b73b0e70246b0e7f825482c158f2ed047d2a5. The portable `${rootDir}${path.sep}locales/fr/messages.po` spelling is retained, and the assertion now uses `toStrictEqual` with the complete diagnostic object: `code`, exact `message`, `catalogPath`, and `locale`. Its strict shape also rejects an unexpected `id` or any other extra field. Evidence: focused core-node fixture passed, then the full package/workflow JS/TS test suite passed.

crates/palamedes/src/transform/visitor.rs
```diff
@@ -74,6 +80,37 @@ impl<'a> TransformVisitor<'a> {
...
+    pub(super) fn rebind_surviving_trans(&mut self, semantic: &Semantic<'_>) {
+        let mut aliases = HashMap::<SymbolId, String>::new();
+        let mut reserved = self.imports.used_identifier_names.clone();
+        for (symbol_id, module, replacement_index) in self.trans_replacements.clone() {
+            let needs_alias = self.imports.has_surviving_reference(
+                semantic,
+                symbol_id,
+                &self.consumed_binding_ranges,
+            ) || (!self.imports.has_reusable_trans_import(&module)
+                && self.imports.has_other_binding_named("Trans", symbol_id));
```
swernerx: [P1] Check `Trans` reuse at the replacement scope

`has_reusable_trans_import` is module-global, so it suppresses the collision guard even when that import is shadowed where the generated JSX lands. This valid input reproduces it: `import { Trans as MacroTrans } from "@palamedes/react/macro"; import { Trans } from "@palamedes/react/compiled"; function Example() { const Trans = () => "authored-shadow"; return <MacroTrans>Hello</MacroTrans>; }`. The public transform emits `<Trans id=...>` inside `Example`; executing the transpiled module produces an element whose type is the authored function and returns `authored-shadow`, not the compiled component. Use the shared semantic facts to prove that bare `Trans` resolves to the exact reusable import at each replacement scope, otherwise emit the collision-free alias, and add this nested-shadow regression.

swernerx (commented):
Verdict: changes requested

All four previous findings are fixed at `569b73b0e70246b0e7f825482c158f2ed047d2a5`: the same-binding type query keeps its macro import, authored `Trans` collisions receive a compiled alias, type-only and wrong-export compiled imports are not reused, and the Windows diagnostic assertion is now strict with the complete object. The exact mixed-use `t` module also executes through the public artifact and reaches the intentional runtime-stub error instead of `ReferenceError`.

One P1 remains in the surrounding `Trans` reuse path. A compiled import is treated as globally reusable even where a nested authored `Trans` shadows it, so the transform silently routes generated JSX to the wrong component. I left the executable repro inline. Focused transform tests, Rust format/Clippy, the rebuilt public package, core-node tests, JS/TS format/lint, and the pinned-toolchain type/declaration checks passed. GitHub CI is terminal at six passes and one macOS CLI signal-test failure (`130 !== 0`) outside this patch's changed files.

crates/palamedes/src/transform/visitor.rs
swernerx: Fixed in d523ab48e88e64e5c976c92c9a759a99b8d8bb8d. Reuse is now proven per macro replacement: shared OXC reference scope facts resolve bare `Trans` at that site and it is reused only when it is the exact compiled-import symbol; otherwise the transform emits a collision-free alias.

Added the nested authored-shadow regression at the native transform layer and an executed core-node public-output regression. Evidence: `cargo +1.95 test -p palamedes transform::tests:: --locked` (98 passed); rebuilt package then focused core-node execution test (1 passed); explicit `packages/core-node/dist/index.mjs` replay returned `compiled-component`, not `authored-shadow`; Rust 1.95 and stable workspace format/Clippy/tests plus JS build, native-types, types, format, lint, and tests are green.

swernerx (commented):
Verdict: clear

The round-2 fix is correct at `d523ab48e88e64e5c976c92c9a759a99b8d8bb8d`. `Trans` reuse is now decided from the exact OXC reference scope and compiled-import symbol at each replacement site. The rebuilt public artifact executed the original nested-shadow case through the generated alias and returned the compiled component; a mixed module also reused bare `Trans` at unshadowed and sibling-only sites without routing the shadowed site incorrectly.

I rechecked the four earlier fixes and the complete cumulative diff. The mixed `t` runtime path remains bound and reaches the intentional macro stub, same-binding type queries retain their import, authored and type/value import collisions remain safe, and the Windows diagnostic assertion stays exact, strict, and test-only. No new findings surfaced.

Local evidence is green: 98 focused native transform tests, Rust 1.95 format and workspace Clippy, the rebuilt package, 29 core-node tests plus its package test, native declarations and TypeScript checks, lint, formatting, deterministic source maps, and executed public-output fixtures. GitHub CI is terminal with all 7 checks passing.

PR #623, opened 2026-08-08
feat(waku): add request-scoped i18n for server actions

### LINKED ISSUE #554: Spike and implement request-scoped i18n for Waku server actions
## Context

Follow-up to #531. The Next.js phase was implemented in #543, but Waku has different runtime extension points and should not inherit compile-time injection without evidence that it is necessary.

Waku supports server actions alongside request middleware and interceptors. We need to verify whether a supported runtime hook encloses the complete server-action invocation, including parameter binding and the full asynchronous/streaming lifetime.

## Goal

Allow an application to configure request-scoped Palamedes i18n once for Waku server actions instead of calling an initializer in every action body.

Prefer a runtime wrapper over source transformation when Waku exposes a reliable end-to-end request boundary. Applications continue to own locale negotiation and catalog loading; Palamedes owns placing the activation around the action invocation.

## Investigation and direction

- Identify the stable Waku interceptor or middleware API suitable for server actions.
- Prove whether the scope begins before action parameter/default evaluation.
- Prove whether the scope remains active through asynchronous helpers and streaming.
- Verify access to the original request headers and cookies for locale resolution.
- If no runtime hook covers the full invocation, document the gap and evaluate reuse of the directive-aware Server Function transform introduced for Next.js.
- If body injection is required, carry forward the parameter-initializer safety diagnostic from #531 rather than silently initializing too late.

## Acceptance criteria

- [ ] A real client-to-server Waku action initializes Palamedes i18n without a per-action initializer call.
- [ ] Direct macros and macros in synchronous, asynchronous, and cross-module helpers use the active request locale.
- [ ] The chosen boundary is proven to cover parameter binding and the complete asynchronous invocation, or the limitation and transform fallback are documented.
- [ ] Concurrent requests with different locales remain isolated.
- [ ] Locale resolution can inspect the original request headers and cookies.
- [ ] Initializer and catalog-loading failures surface as clear server errors.
- [ ] The integration is opt-in and existing applications remain unchanged.
- [ ] Runtime, streaming, and Edge/AsyncLocalStorage limitations are documented.
- [ ] The final decision between runtime wrapping and compile-time instrumentation is recorded with integration tests.

swernerx:

Closes #554

## Summary
- Add the opt-in `@palamedes/waku` runtime interceptor and public package surface.
- Register it in the production-built Waku cookie fixture, with direct macros plus sync, async, cross-module, and default-parameter server-action proof.
- Cover original request headers/cookies, concurrency, nested/error restoration, initialization failures, and stream callbacks; document Waku beta, Node AsyncLocalStorage, Edge, detached-work, and middleware-ordering limits.

## Validation
- `CI=true pnpm --filter @palamedes/example-waku-cookie build`
- `CI=true pnpm verify:examples:smoke -- --id waku-cookie`
- `CI=true pnpm verify:examples:browser -- --id waku-cookie`
- `CI=true pnpm --filter @palamedes/site build`

## Boundary evidence
The pinned Waku dispatcher invokes `fn(...args)` inside its handler interceptor, so the scope begins before action default-parameter evaluation and spans the awaited action. Waku creates response streams within that scope; Node AsyncLocalStorage preserves context for stream callbacks created by the handler.

No transform was added.


packages/waku/package.json
```diff
@@ -0,0 +1,59 @@
...
+    "directory": "packages/waku"
+  },
+  "bugs": "https://github.com/sebastian-software/palamedes/issues",
+  "license": "MIT",
+  "type": "module",
+  "exports": {
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.mjs",
+      "require": "./dist/index.cjs"
```
swernerx: [P1] Drop the unusable CommonJS contract

This package advertises a require branch and main entry and emits dist/index.cjs, but an actual tarball consumer with waku@1.0.0-beta.8 cannot require("@palamedes/waku"): the CJS file requires waku/router/server and fails immediately in the peer RSC/ESM graph (both with ordinary Node conditions and with react-server enabled). The published-types gate only includes this adapter in its ESM fixture, so it misses the broken advertised path. Please make this adapter honestly ESM-only by removing the require/main path and CJS emission/artifacts, or provide a genuinely loadable CommonJS contract, and add a packed runtime consumer assertion.

packages/waku/README.md
```diff
@@ -0,0 +1,68 @@
...
+an error beginning `Palamedes Waku i18n initialization failed`. Waku's default
+production handler returns its generic 500 response for uncaught errors; keep
+server logs available to retain the error cause.
+
+## Streaming and runtime limits
+
+The interceptor awaits Waku's handler promise. Waku creates the response
+`ReadableStream` inside that scope, so Node's `AsyncLocalStorage` retains the
+active locale in stream callbacks created by the handler. The scope itself is
+restored after the response completes for the caller that initiated it.
```
swernerx: [P2] Restore caller scope at handler settlement

The scope is restored for the caller when the awaited next() promise resolves and returns the response, not when the response body finishes streaming. Stream callbacks created inside the scope can retain their inherited locale, but the caller is already restored while the body is consumed. Please correct this sentence and the corresponding API-doc wording to distinguish handler-promise settlement from stream completion; the current text overstates the request ownership lifetime.

swernerx (commented):
Verdict: changes requested

The request-scoped runtime behavior and real Waku fixtures are sound, but two public-contract/documentation issues remain; both are detailed inline.

Validated at exact head ba5b593b2506f38e71433cb07d420e60971b5b63: 7/7 GitHub checks are green; the production Waku build, deterministic server-locale smoke, and browser action fixture pass; focused package/declaration tests, packed-artifact inspection, the full repository agent gate, example-matrix gate, and production docs build pass. The packed consumer exposes the advertised CommonJS failure described inline.

packages/waku/package.json
swernerx: Implemented in 5156d85c93ccd7b191b730b2f4746601d908f66e.

- Removed `require`, `main`, and CJS emission/artifacts; the export map now exposes only ESM plus declarations.
- Added a real packed-tarball consumer test that packs runtime and Waku, installs exact `waku@1.0.0-beta.8`, dynamically imports `@palamedes/waku`, asserts the ESM-only manifest/artifacts, and verifies CommonJS `require()` rejects with `ERR_PACKAGE_PATH_NOT_EXPORTED` naming `@palamedes/waku`. It uses literal process argument paths, including on Windows.
- Evidence: `CI=true pnpm --filter @palamedes/waku test` (unit + packed consumer), `CI=true pnpm verify:examples:smoke -- --id waku-cookie`, and `RUSTUP_TOOLCHAIN=stable CI=true pnpm agent:check` all passed.

packages/waku/README.md
swernerx: Implemented in 5156d85c93ccd7b191b730b2f4746601d908f66e.

The README and API documentation now state the precise boundary: the caller scope is restored when Waku’s awaited handler promise settles and returns its `Response`, before body consumption. Stream callbacks created inside the scope may retain inherited Node `AsyncLocalStorage` context while consumed later, but they do not extend caller/request ownership.

Evidence: `CI=true pnpm --filter @palamedes/site build` and `RUSTUP_TOOLCHAIN=stable CI=true pnpm agent:check` passed.

swernerx (commented):
Verdict: clear

Both previous findings are fully addressed at exact head 5156d85c93ccd7b191b730b2f4746601d908f66e. The packed @palamedes/waku artifact is honestly ESM-only, dynamically imports with real waku@1.0.0-beta.8, and rejects require() at its own export map. The lazy request-accessor import keeps plain adapter import outside the Waku RSC graph while the production interceptor resolves the original Request inside Waku before initialization and next(). README, API, stability, site, and LLMS wording now distinguish caller restoration at handler-promise settlement from inherited async-resource context.

Independent validation passed: focused unit and packed-consumer tests, full repository build/release/declaration/LLMS/lint/type/Rust gate, production Waku deterministic concurrency smoke, browser-to-server action fixture, example inventory, and production docs site. GitHub CI is terminal 7/7 green at the exact head. No new findings.

PR #624, opened 2026-08-08
feat: add React Router RSC i18n scope

### LINKED ISSUE #565: Spike and implement request-scoped i18n for React Router RSC Server Functions
## Context

Follow-up to #531 and parallel to #554.

Palamedes already supports React Router Framework Mode route actions (`export async function action`): the browser-verified examples resolve and activate the request locale inside each action and return localized output. This issue is not about those route actions.

React Router now also exposes React Server Functions via `"use server"` in its experimental RSC Framework and Data Modes. RSC Framework Mode uses the unstable React Router RSC Vite plugin together with `@vitejs/plugin-rsc`, and supports custom RSC entry files. Palamedes currently has no React Router RSC adapter or example that initializes request-scoped i18n for these independent Server Function invocations.

References:

- https://reactrouter.com/how-to/react-server-components
- https://react.dev/reference/rsc/use-server
- #531 — framework-neutral request-scoped Server Function initialization
- #554 — corresponding Waku spike

## Goal

Allow an application to configure request-scoped Palamedes i18n once for React Router RSC Server Functions instead of calling an initializer in every `"use server"` function body.

Prefer a runtime request wrapper over source transformation when React Router's RSC entry or request APIs expose a reliable end-to-end boundary. Applications continue to own locale negotiation and catalog policy; Palamedes owns placing activation and catalog loading around the Server Function invocation.

Treat React Router RSC support as experimental and opt-in while the upstream APIs remain unstable.

## Investigation and direction

- Start with RSC Framework Mode and a minimal production-built example using `unstable_reactRouterRSC` plus `@vitejs/plugin-rsc`.
- Verify whether a custom `app/entry.rsc.ts(x)` wrapper around the RSC request handler encloses Server Function dispatch before argument/default-parameter evaluation.
- Verify that the scope remains active through asynchronous helpers, RSC rendering, revalidation, and streaming.
- Verify access to the original `Request`, including headers and cookies, for locale resolution.
- Determine whether the existing Vite graph-split server message registration/loading path can supply only the active locale's fragments for evaluated Server Function modules.
- Prove concurrent requests with different locales remain isolated across the RSC and SSR module graphs.
- Record whether the same integration contract can cover RSC Data Mode or whether Data Mode needs a separate adapter/documented setup.
- If no runtime hook covers the full invocation, document the gap and evaluate reuse of the directive-aware Server Function transform introduced for Next.js.
- If body injection is required, carry forward the parameter-initializer safety diagnostic from #531 rather than silently initializing too late.

## Acceptance criteria

- [ ] A real client-to-server React Router RSC Server Function initializes Palamedes i18n without a per-function initializer call.
- [ ] The proof uses `"use server"`, not a classic route `action`.
- [ ] Direct macros and macros in synchronous, asynchronous, and cross-module helpers use the active request locale.
- [ ] The chosen boundary is proven to cover parameter binding and the complete asynchronous/streaming invocation, or the limitation and transform fallback are documented.
- [ ] Concurrent Server Function requests with different locales remain isolated.
- [ ] Locale resolution can inspect the original request headers and cookies.
- [ ] Only the active locale's required server message fragments are loaded when graph splitting is enabled, or the initial catalog-loading limitation is explicitly documented.
- [ ] Server Function results and the subsequent React Router revalidation render with the same locale.
- [ ] Initializer and catalog-loading failures surface as clear server errors.
- [ ] The integration is opt-in and existing non-RSC React Router applications remain unchanged.
- [ ] RSC Framework Mode is covered by a production integration test; RSC Data Mode support or non-support is documented.
- [ ] Runtime, streaming, deployment-runtime, and AsyncLocalStorage limitations are documented.
- [ ] The final decision between entry/request wrapping and compile-time instrumentation is recorded with integration tests.

swernerx:
Closes #565

## Summary

- Adds the opt-in `@palamedes/react-router-rsc` adapter and a custom `entry.rsc.tsx` request boundary.
- Adds a production-built React Router RSC fixture with a Client Component invoking a real top-level `"use server"` Server Function.
- Wires package publishing, release version propagation, declarations, API docs, TypeDoc, LLMS inventory, and a focused fixture verifier.

## Runtime decision

The custom RSC entry wraps React Router 8.3.0's default `entry.rsc` `fetch()` inside `@palamedes/runtime/server` AsyncLocalStorage. This starts before `unstable_matchRSCServerRequest`, covering argument/default-parameter evaluation, dispatch, awaited work, RSC rendering, SSR/revalidation, and stream creation. No directive-aware transform is used.

## Evidence

- The production browser fixture runs concurrent English/German requests, client-to-server Server Function calls, direct/sync/async/cross-module macro paths, default parameters, and post-action revalidation.
- Resolver failure is tested to reject before dispatch.
- The fixture confirms the original request cookie/Accept-Language headers reach the resolver.

## Limitations

- Supports only experimental React Router RSC Framework Mode at `@react-router/dev` / `react-router` 8.3.0 and `@vitejs/plugin-rsc` 0.5.32.
- RSC Data Mode and non-RSC applications are unsupported.
- Load the active locale's server catalog in the resolver: Vite does not expose a safe RSC Server Function module-to-fragment manifest.
- Node-compatible AsyncLocalStorage is required; Edge/Worker runtimes and detached async work need application-specific handling.

## Validation

- `env RUSTUP_TOOLCHAIN=1.95 CI=true pnpm agent:check`
- `CI=true pnpm build:site`
- `CI=true pnpm verify:react-router-rsc`


examples/react-router-rsc-cookie/app/lib/i18n.ts
```diff
@@ -0,0 +1,24 @@
...
+
+import { createI18n } from "@palamedes/core"
+import { messages as deMessages } from "../locales/de.po"
+import { messages as enMessages } from "../locales/en.po"
+
+export type Locale = "en" | "de"
+
+function resolveLocale(request: Request): Locale {
+  const cookie = request.headers.get("cookie") ?? ""
+  if (/(?:^|;\\s*)locale=de(?:;|$)/u.test(cookie)) {
```
swernerx: [P1] Parse whitespace between cookies, not a literal backslash. In a RegExp literal, `\\s*` matches a backslash followed by `s` characters; it does not match the normal space after `;`. Against the built production fixture, `Cookie: locale=de` rendered German but `Cookie: session=abc; locale=de` silently rendered English. Please use the intended whitespace expression (or a cookie parser) and cover the multi-cookie Request so the header/cookie acceptance proof is realistic.

scripts/verify-react-router-rsc.mjs
```diff
@@ -0,0 +1,105 @@
...
+    await Promise.all([
+      deContext.addCookies([{ name: "locale", value: "de", url: baseUrl }]),
+      enContext.addCookies([{ name: "locale", value: "en", url: baseUrl }]),
+    ])
+    const [dePage, enPage] = await Promise.all([deContext.newPage(), enContext.newPage()])
+    await Promise.all([dePage.goto(baseUrl), enPage.goto(baseUrl)])
+
+    await expectText(dePage, "server-rendered-message", "Server-Rendern bestätigte Sprache.")
+    await expectText(enPage, "server-rendered-message", "Server render confirmed locale.")
+    await Promise.all([
```
swernerx, line 99: [P1] Force the two initialized server requests to rendezvous before claiming concurrent isolation. `Promise.all()` starts two browser interactions but does not prove their RSC POST/SSR work overlaps; with the current one-microtask helper, either request may complete before the other dispatches, so a shared global locale could still pass. The repository's `@palamedes/runtime/server-test` barrier exists specifically to prevent this false positive. Please add an in-scope barrier/response marker and require both different-locale requests to reach it across the actual production RSC/SSR path.

packages/react-router-rsc/README.md
```diff
@@ -0,0 +1,122 @@
...
+that work. Initialization failures reject before React Router dispatches a
+Server Function, with an error beginning `Palamedes React Router RSC i18n
+initialization failed`.
+
+## Current limits
+
+- The adapter relies on Node's `AsyncLocalStorage` through
+  `@palamedes/runtime/server`. Do not use it in Edge or Worker runtimes unless
+  that runtime's Node-compatible async context propagation has been verified.
+- Detached work started after a Server Function resolves is outside the request
```
swernerx: [P1] Distinguish caller restoration from inherited async-resource lifetime. `AsyncLocalStorage.run()` restores the caller when the awaited dispatch settles, but timers, promises, and stream callbacks created inside that scope retain its store even when they execute later. I verified the packed package: after `run()` resolved the caller's `getI18n()` threw, while a timer scheduled inside still observed `de`. Please correct this statement and the corresponding ADR/LLMS blanket `detached work` claims to describe newly initiated out-of-scope work separately from resources inherited from inside the request.

swernerx (commented):
Verdict: changes requested

Reviewed exact head `202e74a2e9e166141b301266867ec6dda70a8c24` from a fresh detached worktree; both remote refs still match, and GitHub is 7/7 green with MERGEABLE/CLEAN state.

The core runtime boundary is otherwise sound: the production fixture emits a real client Server Function reference and POST dispatch, the pinned React Router 8.3.0 entry places argument decoding/default evaluation, action execution, awaited work, revalidation, RSC response creation, and SSR beneath the scope, and the browser proof passes direct/sync/async/cross-module/default-parameter/post-action localization. Packed ESM/CJS consumption, declarations, release/LLMS inventory, site prerendering, the complete JS/Rust gate, no client leakage, and the test-only Windows normalization also passed.

Three acceptance gaps remain inline: normal multi-cookie locale resolution is broken, the production verifier does not force concurrent requests to overlap, and the detached-work documentation contradicts the packed runtime's actual AsyncLocalStorage inheritance.

examples/react-router-rsc-cookie/app/lib/i18n.ts
swernerx: Fixed in ca9ed9d2b2b3e1d1829d113011bb0275a941b730. The cookie matcher now uses \s* (one whitespace expression), and the production-built verifier sends Cookie: session=production-proof; locale=de and requires German server-rendered output. Red: this verifier failed against the prior literal-backslash matcher; green: CI=true pnpm verify:react-router-rsc passes.

scripts/verify-react-router-rsc.mjs
swernerx: Fixed in ca9ed9d2b2b3e1d1829d113011bb0275a941b730. The custom production entry now awaits the opt-in runtime rendezvous after locale activation and marks the actual default-entry POST response. The browser verifier injects one unique barrier header into both real client Server Function POSTs and requires matching response markers, so it fails unless both requests overlap. CI=true pnpm verify:react-router-rsc passes.

packages/react-router-rsc/README.md
swernerx: Fixed in ca9ed9d2b2b3e1d1829d113011bb0275a941b730. README, API docs, runtime docs, ADR, and LLMS now distinguish caller restoration from inherited async-resource lifetime: resources created inside the scope retain its store when later executed; separately initiated work without inherited context has none. Site build and CI=true pnpm agent:check pass.

swernerx (commented):
Verdict: clear

Reviewed exact head `ca9ed9d2b2b3e1d1829d113011bb0275a941b730` in a fresh detached worktree.

The three repair items are verified:

- The production verifier sends `Cookie: session=production-proof; locale=de`, and the built server renders the required German marker. The corrected cookie expression accepts the normal whitespace between cookie pairs.
- Two real client-to-server Server Function POSTs carry the same unique rendezvous ID, enter the barrier only after their request-local locale is active, and must return matching response markers. The production browser fixture passed. As a negative control, one arrival timed out after the five-second guard, while a matched pair released, so sequential execution cannot satisfy the proof.
- README, API/runtime docs, ADR, and LLMS now consistently distinguish caller restoration when the awaited entry settles from the later inheritance retained by async resources created inside the scope; separately initiated work without inherited context is documented as requiring explicit data or a new scope. Streaming/response lifecycle wording is consistent with that model.

No fresh regression, weakened Windows assertion, client leakage, test hook activation in normal production, or package/documentation inconsistency found. The generated client graph contains no server-scope, barrier, or `async_hooks` symbols; the server rendezvous is inert unless both its environment opt-in and per-request header are present.

Independent gates passed: focused adapter tests and declarations, example matrix, production React Router RSC build/browser/concurrency fixture, deliberate single-arrival timeout control, packed artifact inspection, full `agent:check` (release, published declarations, LLMS, format, lint, JS/TS tests/types, Rust fmt/clippy/workspace tests), full production site build, and diff hygiene. GitHub reports 7/7 successful checks with `CLEAN` / `MERGEABLE` state at the exact head.

PR #647, opened 2026-08-09
fix(core): validate singular translation patch ICU

### LINKED ISSUE #628: fix(core): singular patch values are never ICU-validated — invalid ICU is silently persisted as "applied"
## Problem

The #569 fix validates plural-branch ICU pre-write (`translation.invalid_icu`, per-patch id), but singular values skip ICU parsing entirely:

- `validate_patch_translation` (`crates/palamedes/src/translation_candidates.rs:932-937`) only checks non-emptiness for `Singular`
- `validate_plural_translation_branches` (`:971-980`) returns `Vec::new()` for non-plural values
- `render_translation_value` (`:1068-1070`) passes the singular string through verbatim
- ferrocat's conversion path does not parse msgstr ICU (verified in ferrocat-po 3.4.2 `conversion.rs:53-73`)

So a singular patch like `"Hallo {name"` for source `"Hello {name}"` is written to disk and reported `Applied`.

## Impact

The exact same class of provider/LLM error that #569 now rejects for plural branches silently corrupts the catalog for singular messages — by far the more common shape — and only surfaces later as `compile.invalid_icu_message` or `pmds audit` `icu.syntax`, far from the patch operation that caused it.

## Direction

Run singular values through the same pre-load ICU validation and emit `translation.invalid_icu` with the patch id. Use the policy-aware parse that compile/audit use (`IcuSyntaxPolicy::RuntimeLiteralApostrophes`) so valid apostrophe forms are not rejected — and consider aligning the plural-branch validation (currently plain `parse_icu`) to that policy at the same time.

*Filed from the 2026-08-08 re-audit.*

swernerx:
## Summary

- Validate singular translation patch values with the runtime ICU syntax policy before any catalog write.
- Keep plural branch validation on the same policy while preserving invalid-ICU diagnostics.
- Add regression coverage for invalid singular ICU and valid literal apostrophes.

## Why

Singular patches could previously persist malformed ICU and report success; they now reject with `translation.invalid_icu` and the patch ID.

## Validation

- `cargo fmt --all --check`
- `rustup run 1.95 cargo test --workspace --locked`
- `rustup run 1.95 cargo clippy --workspace --all-targets --locked -- -D warnings`

## Follow-up

None.

Fixes #628

### DIFF
--- a/crates/palamedes/src/translation_candidates.rs
+++ b/crates/palamedes/src/translation_candidates.rs
@@ -427,7 +427,7 @@ where
             continue;
         }
 
-        for message in validate_plural_translation_branches(&patch.translation) {
+        for message in validate_translation_icu(&patch.translation) {
             rejected.insert(index);
             diagnostics.push(workflow_diagnostic(
                 "translation.invalid_icu",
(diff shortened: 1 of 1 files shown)


swernerx:
Closing as superseded: #628 was resolved and merged by #649 with green CI. This duplicate PR must not be merged independently.

